PR #136, opened 2013-09-09
Implemented different logLevel per each logger

bpoplauschi:
Different log levels per loggers can be very useful. In all the projects I used CocoaLumberjack, I wanted to have different levels for each XCode configuration (Debug vsAdHoc vs AppStore), but also different for each logger.
Potential reasons:
- TTY logs are not needed on AppStore builds
- ASL are slow (even with the improvement from CocoaLumberjack), so I tend to set Warning level only for those in AppStore builds, as where the file logging will use a more permissive setting (like info)
- if you use a custom logger (like I do for Crashlytics logs), chance is you want a different log level there too

### DIFF
--- a/Lumberjack/DDASLLogger.m
+++ b/Lumberjack/DDASLLogger.m
@@ -64,6 +64,10 @@ static DDASLLogger *sharedInstance;
 
 - (void)logMessage:(DDLogMessage *)logMessage
 {
+    if (![self shouldLogMessage:logMessage]) {
+        return;
+    }
+    
 	NSString *logMsg = logMessage->logMsg;
 	
 	if (formatter)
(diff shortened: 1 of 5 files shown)


arrb:
Hi, did you guys ever add setLogLevel? I am trying to change the level of a log once a user clicks a button, I've looked everywhere and I can't find the answer..

bpoplauschi:
@arrb we didn't add `setLogLevel`. Are you interested in trying it in a pull request?

arrb:
Hi, the default of my logs is .Info, then I want to add an option (Switch to on or off) which will allow the user look at the log on a debug level, then switch it back to info. Is there a way that could be possible?

PR #353, opened 2014-09-30
Better application name matching in log file names.

firesteps:
"AppLite ..." will not match when application name is "App".

### DIFF
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -269,7 +269,7 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
  * You can change it by overriding newLogFileName and isLogFile: methods.
  **/
 - (BOOL)isLogFile:(NSString *)fileName {
-    NSString *appName = [self applicationName];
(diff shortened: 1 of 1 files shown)


bpoplauschi:
There is no reason in keeping this open.
About keeping open issues, it seems that we do have something to loose here. 
CocoaPods is computing a score for each library and they use the score for ranking and to visually recommend libraries. https://cocoapods.org/pods/CocoaLumberjack/quality. Version 2.0.3 had a score of 91. I managed to increase that to 111 by: adding a changelog, adding documentation via comments (http://cocoadocs.org/docsets/CocoaLumberjack/) - increased from 33% to 70%, improving our readme and the most important: reducing our number of open issues to under 50. See details about the score computation here: https://guides.cocoapods.org/making/quality-indexes. Bottom line is they consider a project with more than 50 open items as abandoned and I kind of agree. As we've both seen, there are periods when people just add issues over issues end we don't have time to close them, so they pile up. So I suggest we keep them to a minimum. This is the way Mattt Thompson was running things on AFNetworking, never more than 20 open issues, most of the times under 10. He was cleaning them up weekly and closing things that were not problems or features he didn't want. Now look at AFN after he has passed it to some other maintainers. 70+ issues.

PR #430, opened 2014-12-30
Updating naming convention

seltzered:
Changing from legacy LOG_LEVEL_VERBOSE to Lumberjack 2.x DDLogLevelVerbose naming


rivera-ernesto:
Thank you!

PR #436, opened 2015-01-09
Add pre-packaged formatter that includes file name and line number. 

jasperblues:
* Save time setting up CocoaLumberjack
* Serve as documentation on how to create own custom format.

### DIFF
--- a/Classes/Formatters/DDFileAndLineNumberFormatter.h
+++ b/Classes/Formatters/DDFileAndLineNumberFormatter.h
@@ -1,0 +1,20 @@
+// Software License Agreement (BSD License)
+//
+// Copyright (c) 2010-2014, Deusty, LLC
+// All rights reserved.
+//
+// Redistribution and use of this software in source and binary forms,
+// with or without modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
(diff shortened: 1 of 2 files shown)


bpoplauschi:
This should probably be set as an example rather than included in the library, IMO.

jasperblues:
@bpoplauschi Its up to you. I presented my justification in #403. ie, since its so commonly asked for, why make it necessary to read the docs on creating a formatter, implement one, plug it in,  and end up copy/pasting it between projects.

rivera-ernesto:
I think we can add those as long as they are not automatically imported by the main header files.

Also as @bpoplauschi says the formatter should be demonstrated somewhere to make it easy to test. To avoid a big number of bum demos we should probably add it to one of the existing formatter-related demos. Could you do that?

bpoplauschi:
@jasperblues could you update the PR with our latest comments?

bpoplauschi:
@jasperblues any updates?

jasperblues:
Let's close this PR. I'm no longer interested in this feature. At the time I thought it might be useful for others. Not sure if this is the case, but presently I don't have sufficient time to work on this.

bpoplauschi:
Ok. Thanks anyway.

PR #443, opened 2015-01-28
Declare automaticallyAppendNewlineForCustomFormatters properties as nonatomic

ferrous26:
They were implicitly atomic and do not need to be. Noticed this in a project where the relevant warnings were enabled.


Classes/DDTTYLogger.m
```diff
@@ -1247,7 +1247,7 @@ - (void)logMessage:(DDLogMessage *)logMessage {
             v[2].iov_base = (char *)msg;
             v[2].iov_len = msgLen;
 
-            if (_automaticallyAppendNewlineForCustomFormatters) {
+            if (iovec_len == 5) {
```
rivera-ernesto: So `_automaticallyAppendNewlineForCustomFormatters` is now simply ignored?

ferrous26: It is not ignored, `_automaticallyAppendNewlineForCustomFormatters` was checked a few lines above the diff context shown here. `_automaticallyAppendNewlineForCustomFormatters` was used to decide how large the `io_vec` should be. 

Without this change, checking `_automaticallyAppendNewlineForCustomFormatters` again here causes a potential race condition if the property is changed from `@NO` to `@YES`.

rivera-ernesto: I see.

PR #444, opened 2015-01-29
Enable codesign for iOS device framework builds

drance:
The CocoaLumberjack framework target in /Framework/Desktop has supported platform provisions for the iOS Simulator, but when you try to build it for a device (either directly or via a separate workspace target that links CocoaLumberjack.framework), it hits a dependency failure:

"CodeSign error: code signing is required for product type 'Framework' in SDK 'iOS 8.1'"

This is because the CocoaLumberjack target has no codesign settings for iOS, as it appears to have started life as a Mac target. I was able to remedy this by adding the appropriate codesign settings to the two build configs manually. The framework now builds for iOS Device, and can build and run larger projects linking CocoaLumberjack.framework on provisioned hardware.


drance:
First commit was too aggressive and turned code signing on for everything, including Mac. It also neglected CocoaLumberjackSwift, which suffered from a similar codesign error. I'm now only applying the sdk-iphoneos subsetting, and graduated it to the project level so both CocoaLumberjack and CocoaLumberjackSwift can benefit.

screen shot 2015-01-29 at 8 59 20 am

rivera-ernesto:
Thanks!

PR #446, opened 2015-02-04
Simplify framework and add Carthage support.

mackross:
Framework is in root so that Carthage picks it up. Added duplicate targets for iOS & iOS Swift.


rivera-ernesto:
Funny, just took a look at Carthage yesterday.

Could you add a Carthage Travis test to make sure it works (now and in the future)?

Framework/Lumberjack/CocoaLumberjack.h
```diff
@@ -0,0 +1,39 @@
...
+//
+//
+
+#import <Foundation/Foundation.h>
+
+//! Project version number for CocoaLumberjack.
+FOUNDATION_EXPORT double CocoaLumberjackVersionNumber;
+
+//! Project version string for CocoaLumberjack.
+FOUNDATION_EXPORT const unsigned char CocoaLumberjackVersionString[];
```
rivera-ernesto: We shouldn't duplicate this file.

Framework/Mobile/LibTest/AppDelegate.m
```diff
@@ -1,41 +0,0 @@
-#import "AppDelegate.h"
-#import "ViewController.h"
-
-#import <CocoaLumberjack/CocoaLumberjack.h>
-#import "Formatter.h"
```
rivera-ernesto: It's a good idea to remove this from the framework, but we should move it to Demos.

Framework/Lumberjack/CocoaLumberjack.h
happy-deploy: This one actually uses <CocaLumberjack/*> imports instead of “" as its the framework umbrella header (won’t work for static libs). We could rename either this header or the original and then have this import the other header? What do you think?

mackross: This one actually uses <CocaLumberjack/*> imports instead of “" as its the framework umbrella header (won’t work for static libs). We could rename either this header or the original and then have this import the other header? What do you think?

Framework/Mobile/LibTest/AppDelegate.m
mackross: I actually remade this target and a few of the others inside the framework folder. I used the same project for both OS X & iOS and set it on a per target basis. I'm not sure about removing them, as they do serve to check that the dynamic frameworks & static libs can be built and integrated into a project but could be worth moving them to demos. At the moment all the demo projects use pods? Perhaps they should use the built frameworks?

mackross:
I'll add a travis Carthage build asap.

Framework/Mobile/LibTest/AppDelegate.m
rivera-ernesto: I think demos can achieve the same goal outside the project, plus moving it out keeps the project cleaner once you only need the framework.

Also having some non-CocoaPods demos would be something good. Not sure if they can use Carthage, if so that could be the extra Travis build step.

Framework/Lumberjack/CocoaLumberjack.h
rivera-ernesto: Not sure about what prevents it from being a shared header between CocoaPods, manual importas and framework/Carthage. What is Imports instead of ""?

mackross: Markdown caught my `<>` brackets. The difference is this file has `#import <CocoaLumberjack/_HEADER_>` imports while the other file has `#import "_HEADER_"` imports.

rivera-ernesto: A quick look at UIKit shows that `</>` seems to be preferred. If it's so we should maybe modify the original one and check if that breaks CocoaPods. Not making it breaks Carthage?

apouche:
hey guys I tried using the carthage support on 2.0.0-rc2 but it doesn't want to build the iOS target any idea why that would be ?

bpoplauschi:
@mackross could you please update this? We released 2.0.0 and we are definitely interested in Carthage support

jlawton:
Seems to me that this won't work. Carthage tries to find one Xcode workspace or project in the repo and use that to build the framework. It prioritizes workspaces over projects, but if the first file it tries doesn't contain any Framework targets, it gives up.

In this case, Carthage will pick one of the demo projects, and not find the framework target. Since there are no workspaces in the repo, simply creating one and adding the `LumberjackFramework` project should fix that issue.

guidomb:
Any ideas when this will merged?

rivera-ernesto:
If you want @guidomb you can fork @mackross, fix the conflicts and make another pull request that we can merge.

PR #447, opened 2015-02-05
Added Test Unit Bundle

colemancda:


rivera-ernesto:
Now we need some tests ;)

colemancda:
@rivera-ernesto So can this be merged as a starting point for TDD?

rivera-ernesto:
I would like to have some actual tests, just to make sure it doesn't sit idle (where a pull request to remove it would also make sense!).

bpoplauschi:
@colemancda I appreciate you working on this unit tests setup, but I'm not sure I follow the multitude of changes made here. Could you please detail them? Or take a look at my simpler approach (which I'm not yet sure it works): #474

Framework/Desktop/CocoaLumberjackTests/CocoaLumberjackTests.m
```diff
@@ -0,0 +1,40 @@
...
+}
+
+- (void)testPerformanceExample {
+    // This is an example of a performance test case.
+    [self measureBlock:^{
+        // Put the code you want to measure the time of here.
+    }];
+}
+
+@end
```
rivera-ernesto: Well most of the changes are to Xcode project files. Actual code or tests are still empty, reason why I think we can wait to have some actual tests to add this. Otherwise a pull request to remove unused/emtpy Unit Test would also be valid.

bpoplauschi: So the architecture looks good :)
I will add some valid tests today or tomorrow. I would like to have those before 2.0.0 is released.

rivera-ernesto: Sure. At the time I am not working in a project that uses CL, so you're in the best position to decide when to release it. But adding that can also be done in 2.0.1. In fact most users (CocoaPod users) won't even notice if there are Unit Tests or not.

bpoplauschi: I know, it's just an obsession of mine. I actually want to make sure that when we write the tests, we don't find stuff that needs to be refactored (as it usually happens), so we can indeed include all the breaking changes in 2.0.0

bpoplauschi:
PR replaced by #474. Thanks for the contribution, @colemancda

PR #450, opened 2015-02-06
Copy the message string

MaddTheSane:
Just in case we get passed a mutable string by accident.


rivera-ernesto:
Thank you again @MaddTheSane.

@robbiehanson could you please add him to the group of contributors? Swift contributions are greatly appreciated ;)

bpoplauschi:
Thanks @MaddTheSane

Issue #405, opened 2014-11-27
How to use cocoalumberjack with Swift


CrazyKoala:
Hi.
I've been reading here and there, and I am unsure how to use the framework with swift.

I've seen people aking their own wrappers in swift, but that was quite some months ago.

Following the issues in there, my understanding is that there's an official swift wrapper for using cocoalumberjack.

Having retrieved the lib with cocoapods , the beta 4 from 2.0.0, I still don't see any swift class there, and none of the example projects uses a swift also.

So I'm kinda confused how to make that work with the "official" means.

Any light that you can shed on this ?

Thx

nateuni:
What was the outcome here guys? I am having the same issue.

depl0y:
Yup, got the same question. I know this ticket is closed, but I got the same "No such module" error.

rivera-ernesto:
CocoaPods is very close to releasing Swift support. Does anyone wants to give a try to the [betas](https://github.com/CocoaPods/CocoaPods/blob/master/CHANGELOG.md)?

Sadly I didn't get the time to work on it.

Let's reopen this in the mean time.

oshamsy:
@CrazyKoala @nateuni @depl0y 
I am using current release version of CocoPods (0.35.0) and I got the same issue `compile error : import CocoaLumberjack -> "No such module"`
To get past this for now, I edited the `CocoaLumberjack.swift` file to remove the line:
`import CocoaLumberjack`

Like @rivera-ernesto said, the next version of CocoaPods will have support for Swift (and frameworks) so this issue will be resolved.

Hope this helps.

rivera-ernesto:
What would be possible is to create a 2.1.x Swift branch that uses 0.36.x betas.

oshamsy:
Yes, I think a separate branch is appropriate.

petaren:
I'm also trying to add CocoaLumberjack to a project using Swift. 

I'm using Cocoapods Version 0.36.0.rc.1
My Podfile looks like this:
```
platform :ios, "8.1"

pod "CocoaLumberjack", '2.0.0-rc2'
```

After `pod install` I open the xcworkspace. The `CocoaLumberjack.swift` file is not in the Pods project. So I add it manually from the filesystem. And now I get a *"Nu such module 'CocoaLumberjack'"* on the import CocoaLumberjack` line in the `CocoaLumberjack.swift` file. Same error happens if I try to import it to my own swift files in my own project.

[...]

PR #400, opened 2014-11-18
Add swift to podspec

ValCapri:
Hello,

CocoaPods is going to support swift in 0.36.0.
As stated CocoaPods/CocoaPods#2835 is searching for testing the work in progress.
For testing purpose, I updated your podspec to add the swift file.
And it's pretty well working.

Thanks for all you did,

Lionel Schinckus

Issue #329, opened 2014-08-20
2.0.0 release


rivera-ernesto:
After #269 I was thinking that maybe we should release 2.0.0 as it is. It is as stable as 1.9.x and all it requires is some more documentation about the `#import` updates to do in apps and libraries.

Then I realized that for the first release we should focus in the non-backward compatible changes that we would like to make to the library but couldn't in 1.9.x releases.

For me the biggest is the log level handling:

* Requiring apps to create an arbitrary named variable just to start using CocoaLumberjack is clumsy.
* Third party libraries have no guideline on how to set their own log levels, how to avoid interfering with the apps level and how to let users adjust the levels.

Possible solutions:

##### 1. NBULog-like [dynamic log levels](https://github.com/CyberAgent/iOS-NBULog#dynamic-log-levels) 

While good and working it still requires [defining not conflicting context values and registering](https://github.com/CyberAgent/iOS-NBULog#2-register-a-log-context-and-optional-modules). A more automated way of doing it would be great.

##### 2. [slf4j](http://www.slf4j.org)-like

Abstract the underlying log framework and fallback to `NSLog` or none otherwise.

##### 2. Other?

Related to #225 discussions.

bpoplauschi:
@rivera-ernesto about the items you raised in this issue's header:
> Requiring apps to create an arbitrary named variable just to start using CocoaLumberjack is clumsy.

I created #473, please look at this approach and let me know what you think. I think it's enough, as it would only affect people that don't set `ddLogLevel`.

> Third party libraries have no guideline on how to set their own log levels, how to avoid interfering with the apps level and how to let users adjust the levels.

I really think your initial work in [SL4C](https://github.com/SLF4C/SLF4C) is a great way to handle that.

rivera-ernesto:
I think we're good for a 2.0.0 release. I'm pretty sure that with lots of people upgrading we'll quickly get any fix we may need. for 2.0.1, etc.

Issue #324, opened 2014-08-14
Uniform SDK, system version and available frameworks checks.


rivera-ernesto:
We have a big mess on the way we check for these things and we should uniform and keep only the best and recommended way to check for these things.

Feel free to comment or edit this description directly.

Concretely we need to adopt the best way to:

### Check the platform

a. `#ifdef TARGET_OS_MAC`
b. `#ifdef TARGET_OS_IPHONE`
c. `#if defined(TARGET_OS_IPHONE) || defined(TARGET_IPHONE_SIMULATOR)`

### Check the build time SDK version

a. `#ifdef __IPHONE_8_0`/`__MAC_10_10`
b. `#if __IPHONE_OS_VERSION_MIN_REQUIRED >= 60000`/`#if MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_9`
c. `#if MACOSX_DEPLOYMENT_TARGET < __MAC_10_10 || IOS_DEPLOYMENT_TARGET < __IPHONE_7_0`
### Check the runtime system version

a. `[[[UIDevice currentDevice] systemVersion] floatValue] >= 7.0`
b. `#define SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(v)  ([[[UIDevice currentDevice] systemVersion] compare:v options:NSNumericSearch] != NSOrderedAscending)`
c. `floor(NSFoundationVersionNumber) > NSFoundationVersionNumber_iOS_6_1`

OS X?

### Check if AppKit is available (for CLI tools)

a. `#ifdef NSAppKitVersionNumber10_0`
b. `respondsToSelector:`?
c. `__has_import(<AppKit/NSColor.h>)`

### Check if defined vs check value

a. Prefer `#ifdef VALUE`
b. Prefer directly `#if VALUE`
c. Use both `#if defined(VALUE) && VALUE` as it may be defined but set to `0` (never the case?).

### Other check used in CocoaLumberjack that hopefully won't be longer needed

a. `#ifdef MAC_OS_X_VERSION_10_9`
b. `[NSTimer instancesRespondToSelector:@selector(tolerance)]`
c. `#if !OS_OBJECT_USE_OBJC`
d. `#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 50000`

bpoplauschi:
I think we solved those issues and the issue can be closed.

PR #451, opened 2015-02-09
Fix #433 CocoaLumberjack 2.0.0-rc does not build with Pods

### LINKED ISSUE #433: CocoaLumberjack 2.0.0-rc does not build with Pods
I updated the CocoaLumberjack to 2.0.0-rc from 2.0.0-beta4 and now the project does not build anymore. This are the two errors I get:

```
DDLog.h:344:1: Property with 'retain (or strong)' attribute must be of object type
DDLog.h:539:1: Property with 'retain (or strong)' attribute must be of object type
```

Apparently these lines were changed to point to objects. Is there any specific reason that those properties are marked as strong?

```
@property (nonatomic, DISPATCH_QUEUE_REFERENCE_TYPE, readonly) dispatch_queue_t loggerQueue;
```

bpoplauschi:


rivera-ernesto:
Thanks!

rivera-ernesto:
We could probably release rc2 after this.

bpoplauschi:
We can release rc2, no problem. I want to make a push and release 2.0 stable in a few days. Not sure if we need to wait for CocoaPods to include Swift support.

Before we release, since we only need to make sure the framework doesn't break and all the non-backward compatible changes are made now, I think we still need:
- make sure we have a working stable unit testing solution. #447 has a bit too many changes and I'm not sure how to follow. I will take care of those in a day or two
- Simplify framework and add Carthage support #446 
- make an example of CocoaLumberjack + swift and test it using CocoaPods 0.36 beta. #405 #400 
- regarding #329 do we need to do anything here before a 2.0 goes out?
- what about #324 ?
- #225 is ongoing, so we don't need to close it
- #53 is a bug which ideally would get fixed, but it's not a showstopper.

rivera-ernesto:
I think that we can release a 2.0.0 as soon as there are no critical bugs and backwards compatibility works. It is a x.0.0 release, so we can always work over it.

Swift could be 2.1.0 and so on.

PR #455, opened 2015-02-13
Fix #433 CocoaLumberjack 2.0.0-rc does not build with Pods

### LINKED ISSUE #433: CocoaLumberjack 2.0.0-rc does not build with Pods

oshamsy:
More details on OS_OBJECT_USE_OBJC  and CocoaPods
http://guides.cocoapods.org/syntax/podspec.html#deployment_target

### DIFF
--- a/Classes/DDLog.h
+++ b/Classes/DDLog.h
@@ -23,7 +23,7 @@
     #import "DDLegacyMacros.h"
 #endif
 
-#if OS_OBJECT_HAVE_OBJC_SUPPORT || __IPHONE_OS_VERSION_MIN_REQUIRED >= 60000
+#if OS_OBJECT_USE_OBJC
     #define DISPATCH_QUEUE_REFERENCE_TYPE strong
 #else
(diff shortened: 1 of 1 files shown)


rivera-ernesto:
Looks ok, let's give it a try.

rivera-ernesto:
If this is really fixed it would be great to release rc2!

oshamsy:
I believe this is fixed. I tested with OS X 10.7 and 10.10, as well as iOS 5.0 and 8.1.

PR #458, opened 2015-02-25
Added swift files to podspec.

petaren:
Just added all swift files in the Classes/ directory to the podspec. Tried it myself with CocoaPods, seems to work fine!

### DIFF
--- a/CocoaLumberjack.podspec
+++ b/CocoaLumberjack.podspec
@@ -31,7 +31,7 @@ Pod::Spec.new do |s|
   end
 
   s.subspec 'Core' do |ss|
-    ss.source_files = 'Classes/DD*.{h,m}'
+    ss.source_files = 'Classes/DD*.{h,m}','Classes/*.swift'
   end
 
   s.subspec 'Extensions' do |ss|


CocoaLumberjack.podspec
```diff
@@ -31,7 +31,7 @@ Pod::Spec.new do |s|
   end
 
   s.subspec 'Core' do |ss|
-    ss.source_files = 'Classes/DD*.{h,m}'
+    ss.source_files = 'Classes/DD*.{h,m}','Classes/*.swift'
```
rivera-ernesto: You can just make it `ss.source_files = 'Classes/DD*.{h,m,swift}'`

rivera-ernesto: You could just add it above as `ss.source_files = 'Classes/CocoaLumberjack.{h,m,swift}'`.

rivera-ernesto:
Could you make the pull request to the new [`swift_support`](https://github.com/CocoaLumberjack/CocoaLumberjack/tree/swift_support) branch?

Also lets start using 0.36 in Travis by using `- gem install cocoapods --pre --quiet` [here](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.travis.yml#L9).

petaren:
@rivera-ernesto Yes, I'll redo it. I got a bold idea, not sure if it's good. Tried it out locally and it works. The idea is to use the `podspec.prepare_command` and have some script remove the swift support from the podspec file if the cocoapods version installed is bellow 0.36. 

How would you feel about that?

rivera-ernesto:
We only support the latest Cocoapods, so probably that won't be needed soon. I would suggest you to only concentrate in 0.36 for Swift and that branch.

On the other hand there are many Travis [errors](https://travis-ci.org/CocoaLumberjack/CocoaLumberjack/builds/52124924) that I wonder if using 0.36.x will solve.

petaren:
Okey, ditched the conditional swift support. Will create new pull-request in a minute.

PR #459, opened 2015-02-25
Swift support

petaren:


rivera-ernesto:
Supersedes #458.

Travis has many errors thought. Maybe the Swift source should move to a non-default subspec, to keep compatibility with projects targeting older iOS versions.

Looks like a good start so I'll merge it so work people can continue working on top of this.

petaren:
@rivera-ernesto I noticed that part of the reason travis fails is because we need to make CocoaPods make frameworks of CocoaLumberjack instead of static libraries. To do so one needs to add the `use_frameworks!` attribute in the `Podfile`. The problem occurs in the Demo projects. But I've noticed that a lot of the Demo projects are targeting old versions of OS X and iOS, such as 10.7 and iOS 5.

rivera-ernesto:
Yes, because we still support older system versions.

If there is no better way to exclude Swift code from pre iOS 7/OS X 10.9 projects we may need to create a Swift subspec as I commented above, and people with targeted newer systems would use something like `pod 'CocoaLumberjack/SwiftSupport'`. Unless again there's a better way to keep backwards/non-Swift compatibility.

Once that fixes Travis we could have a Swift demo project that demonstrates `use_frameworks!`. Which I read somewhere is only iOS 8, which sucks because it iOS 7 users that do support Swift won't be able to use CocoaPods.

PR #461, opened 2015-02-25
fix locale for dateFormatter to provide fixed output date format.

ConfusedVorlon:
If the locale is not specified, then the expected format can be overridden in iOS
by the 'use 24hr clock' setting.

This causes filenames with am/pm format, which then fail the isLogFile
test.

This in turn causes sortedLogFilePaths and friends to return zero
entries.

(this bug also exists in the 1.9 branch)

for more detail, see:
https://developer.apple.com/library/ios/qa/qa1480/_index.html


rivera-ernesto:
Thanks!

PR #462, opened 2015-02-26
Added manual install instructions and fixed sub project

nrbrook:


bpoplauschi:
Thanks @nrbrook

Documentation/GettingStarted.md
```diff
@@ -8,16 +8,31 @@ There are 3 steps to getting started with the logging framework:
...
 
-The main files you need to add are:
+#### CocoaPods
 
--   DDLog (Basis of entire framework)
--   DDASLLogger (sends log statements to Apple System Logger, so they show up on Console.app)
--   DDTTYLogger (sends log statements to Xcode console - if available)
--   DDFileLogger (sends log statements to a file)
+```ruby
+platform :ios, '6.1'
```
rivera-ernesto: Shouldn't it be `'5.0'`?

nrbrook: This was copied from README.md. If it should be 5.0 (I'm not sure) it needs changing in both.

rivera-ernesto:
I think it would be better to change both. Otherwise it gives the impression that iOS 5 is not supported.

PR #464, opened 2015-03-01
New log file is always created at app launch on simulator

benrobjoh:
When an app that is capable of running in the background runs on the simulator, the most recent log file is always archived, regardless of file size or age. A new log file is created on each app launch because `NSFileProtectionKey` is not present.

This pull request checks whether there is a value for `NSFileProtectionKey` in the log file's attributes. If there is no file protection value present, it will not force a new log file to be created.


rivera-ernesto:
Thanks!

PR #466, opened 2015-03-03
Support for deleting all old log files at runtime

Autoc0diq:
Setting maximumNumberOfLogFiles to 0 did not delete all files including current log file.
It is now possible to delete all log files at runtime using this approach:

```
[fileLogger rollLogFileWithCompletionBlock:^{
        // forcing maximum number of log files to 0 will delete old files
        fileLogger.logFileManager.maximumNumberOfLogFiles = 0;

        [fileLogger rollLogFileWithCompletionBlock:^{
            fileLogger.logFileManager.maximumNumberOfLogFiles = MaxNumberOfLogFiles;
        }];
    }];
```

### DIFF
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -194,12 +194,10 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
         }
     }
 
-    if (maxNumLogFiles) {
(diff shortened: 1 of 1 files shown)


rivera-ernesto:
Shouldn't `0` mean no limit instead?

Maybe a `deleteLogFiles` would be more clear?

bpoplauschi:
@Autoc0diq what do you say about @rivera-ernesto 's comment?

PR #473, opened 2015-03-09
Default ddLogLevel set to DDLogLevelVerbose

bpoplauschi:
To avoid 
> Requiring apps to create an arbitrary named variable just to start using CocoaLumberjack is clumsy. 

we can define the global log level to **Verbose** unless it's already defined.

### DIFF
--- a/Classes/DDLogMacros.h
+++ b/Classes/DDLogMacros.h
@@ -24,7 +24,11 @@
  * The constant/variable/method responsible for controlling the current log level.
  **/
 #ifndef LOG_LEVEL_DEF
-    #define LOG_LEVEL_DEF ddLogLevel
+    #ifdef ddLogLevel
+        #define LOG_LEVEL_DEF ddLogLevel
+    #else
+        #define LOG_LEVEL_DEF DDLogLevelVerbose
+    #endif
 #endif
 
 /**


rivera-ernesto:
Good idea!

vercruesse:
For Swift the global log level ist still set to .Warning instead of .Verbose

```swift
public var defaultDebugLevel = DDLogLevel.Warning

public func resetDefaultDebugLevel() {
    defaultDebugLevel = DDLogLevel.Warning
}
```

bpoplauschi:
Should we set it to Verbose, to make it consistent with the Objective-C default one?

rivera-ernesto:
I think so.

Classes/DDLogMacros.h
```diff
@@ -24,7 +24,11 @@
  * The constant/variable/method responsible for controlling the current log level.
  **/
 #ifndef LOG_LEVEL_DEF
-    #define LOG_LEVEL_DEF ddLogLevel
+    #ifdef ddLogLevel
```
rivera-ernesto: Also noticed that this check never works as usually `ddLogLevel` is defined as a variable instead of a preprocessor `#define`.

PR #476, opened 2015-03-13
Cleanup .travis.yml

jPaolantonio:
* Use --project-directory instead of cd


rivera-ernesto:
No need to always close pull requests, you could have also added commits directly here.

PR #482, opened 2015-03-22
Update older syntax to modern subscripting for array access

daniel-beard:
- Several minor style changes for array access


bpoplauschi:
Thanks @daniel-beard

Issue #481, opened 2015-03-20
Legacy Warnings with 2.0.0 and Cocoapods Frameworks 


justin:
I'm using Cocoapods 0.36's framework support. The way it imports by default spills my warnings with notices about legacy macros being imported and ambiguous #defines. 

![screen shot 2015-03-20 at 8 49 56 am](https://cloud.githubusercontent.com/assets/1384/6753769/1b54ae3a-cede-11e4-9051-b215a4ef1994.png)

I've tried defining `DD_LEGACY_MACROS` in my code to get around it, but that doesn't seem to pick up in the right spot. Any suggestions?

rivera-ernesto:
Import `CocoaLumberjack.h` instead of `DDLog.h` (as explained in the warning messages).

justin:
We are using `@import CocoaLumberjack`. The only thing that imports `DDLog.h` file is the umbrella header generated by the Lumberjack pod.

rivera-ernesto:
I see. Could you share a sample project with a sample Podfile to reproduce this?

MaddTheSane:
Question: Does CocoaPods support module map files? I have the module map file exclude `DDLog+LOGV.h` and `DDLegacyMacros.h` from the module map file, making the use of your own macros possible.

MaddTheSane:
Although, looking through Clang's documentation, it seems I may have overlooked some use-cases.

rivera-ernesto:
Merged #483. I guess we can't do much more. Thanks @MaddTheSane

tapi:
Running into this issue myself, looks like the specification of custom module maps is scheduled for [Cocoapods 0.37.0] (https://github.com/CocoaPods/CocoaPods/issues/3003)

FWFabio:
This doesn't seem to fix it with cocoa pods 0.37.1

tapi:
The module map is cached and not updated so you may need to delete the module cache on your machine. We had to do this on all our machines including Jenkins and CircleCi.

fatuhoku:
@tapi How does one go about deleting the said module cache?

tapi:
@fatuhoku `rm -rf ~/Library/Developer/Xcode/DerivedData/ModuleCache` should do the trick

fatuhoku:
Sadly that hasn't helped. The warnings still appear.

tapi:
I was using a custom pod spec that pointed at the module map in question.
Just tried 
[...]

PR #483, opened 2015-03-23
Temporary fix for modules.

MaddTheSane:
Temporary fix for #481.
Note that you will need to declare `ddLogLevel` as a preprocessor macro on the command line.

Hopefully Xcode 6.3 will add support for the `textual header` module syntax.

### DIFF
--- a/Framework/Desktop/Lumberjack/CocoaLumberjack.modulemap
+++ b/Framework/Desktop/Lumberjack/CocoaLumberjack.modulemap
@@ -4,6 +4,11 @@ framework module CocoaLumberjack {
 	export *
 	module * { export * }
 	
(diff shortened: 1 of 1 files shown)


bpoplauschi:
We should bring @MaddTheSane onboard. I remember we asked @robbiehanson to do this. I think everyone would benefit from this

rivera-ernesto:
And from more privileges in the organization so we can do it ourselves next time...
Maybe contacting by email would be better?

bpoplauschi:
@robbiehanson could you add @MaddTheSane to our team?

bpoplauschi:
@robbiehanson kind reminder on the previous comment

Issue #491, opened 2015-04-08
CocoaLumberjack.swift is incompatible with swift 1.2/xcode 6.3


mcmurrym:
The problem is `@autoclosure` has changed: first it is not an attribute and should change from ```#string: @autoclosure``` to  ```@autoclosure #string:```.

Second `@autoclosure` now also implements the behavior of `@noescape` this means in a function like DDLogDebug the logText parameter is not allowed to be passed on to the SwiftLogMacro. It can only be called.

MaddTheSane:
At first I was wondering why I *wasn't* having an issue. Then I realized the App Store version hadn't updated.

I'll look into this when my Xcode updates.

bpoplauschi:
Resolved nicely by @MaddTheSane , integrated into our `swift_support` branch.

PR #492, opened 2015-04-09
Update to Swift 1.2

MaddTheSane:
This updates make CocoaLumberjack compatible with Swift 1.2.

### DIFF
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -50,7 +50,7 @@ public func resetDefaultDebugLevel() {
     defaultDebugLevel = DDLogLevel.Verbose
 }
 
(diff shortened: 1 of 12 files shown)


MaddTheSane:
Fixes #491.

mcmurrym:
I see you are calling logText instead of passing it. If you want to avoid calling it you can change `@autoclosure` to `@autoclosure(escaping)` This will allow the passing of the closure without calling it. :) I would have said this earlier but I just learned about it.

MaddTheSane:
@mcmurrym: Done.

bpoplauschi:
Great, @MaddTheSane. I will merge this into the `swift_support` branch since I'm not sure when we will release the Swift version (2.1).

bpoplauschi:
Merged. I will close this PR.

rivera-ernesto:
Is the last commit failing because Travis is not updated?

bpoplauschi:
It's failing because of some unit test which fails randomly. I need to revisit those. The stuff on the `swift_support` branch looks ok.

PR #495, opened 2015-04-14
Update prefix header name for Swift framework

ferrous26:


rivera-ernesto:
Looks good @MaddTheSane ?

MaddTheSane:
The prefix header file is missing, but since the prefix header is only used if building C/C++/Objective C code, the prefix header isn't needed.

It wouldn't surprise me if Xcode didn't complain about the missing file.

rivera-ernesto:
Right. Merging it.

Issue #488, opened 2015-04-01
DDTTYLoggeer's AppKit detection doesn't work & Path to CLIColor.h include wrong


weissi:
# Problem: AppKit Dependency

CocoaLumberjack's DDTTYLoggeer is in theory able to work without having `AppKit.framework` linked. It tries to check whether AppKit is linked by using `#elif __has_include(<AppKit/NSColor.h>)`. Unfortunately this doesn't work, Apple's clang seems to always find the AppKit includes even if it's using AppKit (verified using Xcode 6.1 and Xcode 6.3 beta 4's clang versions).

The following demo program

```
$ cat /tmp/test.m
#if __has_include(<AppKit/NSColor.h>)
#    error AppKit available
#endif
int main() {
    return 0;
}
```

compiled as

```
$ clang -o /tmp/test /tmp/test.m
/tmp/test.m:2:6: error: AppKit available
#    error AppKit available
     ^
1 error generated.
```

does say the AppKit header files are available but there's no `-framework AppKit` or anything else passed on clang's command line.

Therefore, I think CocoaLumberjack's `#elif __has_include(<AppKit/NSColor.h>)` will always be `true` on OS X and which leads to the `#else` case (where `CLIColor.h` is included) to never be used. That creates a dependency on AppKit in DDLog.

# Problem: Path to `CLIColor.h`

Additionally the code to use the CLIColor version is

```
#else
    // OS X CLI
    #import "CLIColor.h"
    #define DDColor CLIColor
    #define DDMakeColor(r, g, b) [CLIColor colorWithCalibratedRed:(r/255.0f) green:(g/255.0f) blue:(b/255.0f) alpha:1.0f]
#endif
```

but the `#import` to `CLIColor.h` should really be `CLI/CLIColor.h` otherwise it won't find CLIColor.h.

# Possible Solution

 - A preprocessor flag `DD_NO_APPKIT` or `DD_USE_CLICOLOR` or similar that DDTTYLogger would be compiled with CLIColor instead of NSColor to remove the AppKit dependency

rivera-ernesto:
Couldn't it be that AppKit is getting added somewhere else in the project? Build steps, `@import`'s, etc.?

weissi:
@rivera-ernesto I'm just compiling one file (`/tmp/test.m`) from the command line. There's no project or anything else involved.

Try it for yourself and just paste this:

```bash
cat > /tmp/test.m <<EOF
[...]

PR #497, opened 2015-04-18
Issue #488: Support DDLog without AppKit Dependency (#define DD_CLI)

weissi:


rivera-ernesto:
Looks very good, but we need to fix this probably unrelated [Travis error](https://travis-ci.org/CocoaLumberjack/CocoaLumberjack/builds/59026361#L813) first to make sure we don't break anything.

bpoplauschi:
Replaced by #627

PR #498, opened 2015-04-18
Tests for DDLogmessage

pk:
Hi from Cocoapods TestJam,

I'm trying to add some test coverage for the project. Started with the DDLogMessage as it was simple enough.

There are 2 failing tests because I've wrote tests according to the documentation which mentions using DDLogMessageOptions to force copy the file & function strings but the code is actually missing in the implementation.


rivera-ernesto:
This one has some more Travis [errors](https://travis-ci.org/CocoaLumberjack/CocoaLumberjack/builds/59047484#L844) than #499 and #500.

pk:
Hi Ernesto,

this one won't actually build in Travis until the #499 is merged in. I wrote the tests to comply with the documentation in the DDLogMessage so it must fail. Two errors actually where I check file & function attributes copy behaviour.

I think I need to figure out why the #499 is failing first, fix that and then merge this one in.

bpoplauschi:
Thanks @pk for all the unit tests you created and sorry for the delay of adding them into the project.

PR #499, opened 2015-04-18
Fixing DDLogMessage to work according to the documentation

pk:
DDLogMessage constructor describes the behaviour which has not been implemented. I've added the tests and this pull requests implements the behaviour so the test passes.


rivera-ernesto:
Looks good to me. Again we have the same Travis error as in #500. So I guess it's not related to your changes but would be better to fix before merging your other pull requests to ensure we don't break anything.

PR #500, opened 2015-04-18
DDLog core class tests WIP

pk:


rivera-ernesto:
Travis is not happy. But it's only one error [here](https://travis-ci.org/CocoaLumberjack/CocoaLumberjack/builds/59047484#L844).

pk:
I'm actually getting completely different errors when building & testing vanilla master. XCode is complaining that the async tests marks fail after it finished. It is however quite random.

Anyway I'll have a look into it. I'd like to work a bit more on the test for the main class.

Issue #503, opened 2015-04-22
cannot make custom log levels as they do not match the DDLogLevel enum


mmerickel:
In 2.0 it is no longer possible to specify custom log flags/levels because all of the APIs specify `DDLogFlag` and `DDLogLevel` in their signatures and ivars. These APIs should take `NSUInteger` instead to allow for the extra flags that were possible in 1.x.

I've already modified the APIs this way in my fork and I'm happy to contribute it, just ask.

rivera-ernesto:
Please make a pull request so we can take a look.

mmerickel:
I don't have a good answer right now. I have temporarily reverted to the 1.x branch. I think a lot of the changes in 2.0 are nice and enums have certain advantages. However they are not extensible at all.

I make heavy use of the pattern established in https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/Documentation/FineGrainedLogging.md which appears to have no equivalent in the 2.0 branch.

The only good answer I can come up with is to remove the enums and go back to the `#define DDLogFlagVerbose (1 << 5)` concept, and make the api continue to accept `NSUInteger` instead. Possibly defining a `typedef NSUInteger DDLogLevel` and `typedef NSUInteger DDLogFlag` to keep the API clean.

I will submit a PR but unfortunately I think it goes against some of the principals you've placed in the 2.0 branch like using enums and that may be an issue for you.

ffried:
Closing due to lack of activity.

Issue #485, opened 2015-03-29
Which log macro should use if I want to implement my custom log context or level?


inix:
#### 1.The log macro following is mentioned in the doc:
```objc
#define HTTP_LOG_CONTEXT 80

#define HTTPLogError(frmt, ...)     SYNC_LOG_OBJC_MAYBE(httpLogLevel, LOG_FLAG_ERROR,   HTTP_LOG_CONTEXT, frmt, ##__VA_ARGS__)
#define HTTPLogWarn(frmt, ...)     ASYNC_LOG_OBJC_MAYBE(httpLogLevel, LOG_FLAG_WARN,    HTTP_LOG_CONTEXT, frmt, ##__VA_ARGS__)
#define HTTPLogInfo(frmt, ...)     ASYNC_LOG_OBJC_MAYBE(httpLogLevel, LOG_FLAG_INFO,    HTTP_LOG_CONTEXT, frmt, ##__VA_ARGS__)
#define HTTPLogDebug(frmt, ...)    ASYNC_LOG_OBJC_MAYBE(httpLogLevel, LOG_FLAG_DEBUG,   HTTP_LOG_CONTEXT, frmt, ##__VA_ARGS__)
#define HTTPLogVerbose(frmt, ...)  ASYNC_LOG_OBJC_MAYBE(httpLogLevel, LOG_FLAG_VERBOSE, HTTP_LOG_CONTEXT, frmt, ##__VA_ARGS__)
```

Or:
```objc
#define DDLogFatal(frmt, ...)    SYNC_LOG_OBJC_MAYBE(ddLogLevel, LOG_FLAG_FATAL,  0, frmt, ##__VA_ARGS__)
#define DDLogError(frmt, ...)    SYNC_LOG_OBJC_MAYBE(ddLogLevel, LOG_FLAG_ERROR,  0, frmt, ##__VA_ARGS__)
#define DDLogWarn(frmt, ...)    ASYNC_LOG_OBJC_MAYBE(ddLogLevel, LOG_FLAG_WARN,   0, frmt, ##__VA_ARGS__)
#define DDLogNotice(frmt, ...)  ASYNC_LOG_OBJC_MAYBE(ddLogLevel, LOG_FLAG_NOTICE, 0, frmt, ##__VA_ARGS__)
[...]

PR #504, opened 2015-04-27
switch back to macros for log levels

mmerickel:
The log levels should remain extensible, keeping only the defaults
defined by CocoaLumerjack.

I'm happy to work through more details of this later. Most things should be fine except the swift support which I'm less familiar with, however it appears to be making heavy use of the enum features.

Related to #503.

### DIFF
--- a/Classes/DDLog.h
+++ b/Classes/DDLog.h
@@ -97,23 +97,22 @@
  * Documentation/CustomLogLevels.md
  **/
 
-typedef NS_OPTIONS(NSUInteger, DDLogFlag) {
-    DDLogFlagError      = (1 << 0), // 0...00001
-    DDLogFlagWarning    = (1 << 1), // 0...00010
-    DDLogFlagInfo       = (1 << 2), // 0...00100
-    DDLogFlagDebug      = (1 << 3), // 0...01000
-    DDLogFlagVerbose    = (1 << 4)  // 0...10000
-};
+typedef NSUInteger DDLogFlag;
+typedef NSUInteger DDLogLevel;
 
-typedef NS_ENUM(NSUInteger, DDLogLevel) {
-    DDLogLevelOff       = 0,
-    DDLogLevelError     = (DDLogFlagError),                       // 0...00001
-    DDLogLevelWarning   = (DDLogLevelError   | DDLogFlagWarning), // 0...00011
-    DDLogLevelInfo      = (DDLogLevelWarning | DDLogFlagInfo),    // 0...00111
-    DDLogLevelDebug     = (DDLogLevelInfo    | DDLogFlagDebug),   // 0...01111
-    DDLogLevelVerbose   = (DDLogLevelDebug   | DDLogFlagVerbose), // 0...11111
-    DDLogLevelAll       = NSUIntegerMax                           // 1111....11111 (DDLogLevelVerbose plus any other flags)
-};
+#define DDLogFlagError      (1 << 0)
+#define DDLogFlagWarning    (1 << 1)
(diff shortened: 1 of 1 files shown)


MaddTheSane:
The main problem with this method is that it breaks Swift compatibility, as Swift currently can't get those values from macros.

I'll explore another idea when I wake up.

rivera-ernesto:
Besides Swift compatibility, enumerations are cleaner than macros, although not extensible (yet?).

I think we could add many more log levels to the enumerations. By default we can still work with only 5 of them and just group the others. Then people can use the extra levels by writing their own [custom log macros](https://github.com/CocoaLumberjack/CocoaLumberjack/issues/485#issuecomment-87413905).

They could even rename the levels by creating a custom enumeration besad on the default enumeration instead of `NSUInteger`.

mmerickel:
I think I addressed (and agreed with) most of these concerns in the original issue. I just haven't found a good compromise yet as far as what is actually doable. I don't actually want to use macros, but like I said, there are compiler restrictions on what we can do in performant c code. If performance was less of a concern we could easily turn log levels/flags into NSObject classes and make them extensible that way, but then the compiler will never be optimizing out unnecessary log statements.

As far as the "not extensible (yet?)", enums are a part of the c language and not likely to grow an extensibility point anytime soon.

mmerickel:
@MaddTheSane what was your other idea?

MaddTheSane:
I was thinking of merging the `DDLogFlag` and `DDLogLevel` into one `NS_OPTIONS` enum.

mmerickel:
I think the separation of flags and levels is actually a great distinction that I don't see in many other libraries that just use levels.

MaddTheSane:
Agreed.

bpoplauschi:
My 2 cents: I think that using macros was a good practice at the time the library was originally created, but now they are no longer the present. Most users will be satisfied with the log levels we provide by default. In Swift, they can even extend the enums.  At this point where more and more devs are migrating to Swift, the Swift support we offer counts more and more every day.

rivera-ernesto:
Agreed. We should focus on modernizing the code, and maybe even migrating to Swift at some point.

mmerickel:
> Most users will be satisfied with the log levels we provide by default. In Swift, they can even extend the enums.

I really feel like closing this issue directly conflicts with the promises made in the documentation and marketing copy [1]. I personally have extra levels and flags that are not in the library but it's also a "best practice" for embedding lumberjack into a library to avoid using the global logging levels. I currently do this by defining my own levels and context per-library. This was all possible with macros.

I have no problem with moving to swift and modernizing code but I feel like you've dropped an important feature.

[1] https://github.com/CocoaLumberjack/CocoaLumberjack#lumberjack-is-flexible

bpoplauschi:
I agree, but we have to come up with a way to keep it flexible while not using macros

rivera-ernesto:
As @MaddTheSane pointed out you can extend levels by using Swift code.

mmerickel:
How is abandoning objective-c a viable option? It's not exactly a dead language. Is that really the direction of CocoaLumberjack? Since this is closed it would help if some constructive examples were added to #503 or updated in the docs?

rivera-ernesto:
It is not really abandoning Objective-C, it is adopting Objective-C 2.0's `NS_OPTIONS` and `NS_ENUM`, which weren't available when this library started.
In fact Apple has replaced all similar macros with them as well, both to make them cleaner and in preparation for Swift.

mmerickel:
Sure, I was commenting that the only suggestion for accomplishing the flexibility guarantees so far has been to "use swift".

PR #506, opened 2015-04-30
Using type safe DDColor alias instead of #define directive.

MaheshRS:
Using type safe alias for DDColor instead of #define directive. Similarly instead of using #define to define a function that returns UIColor, using static to make it type safe.


rivera-ernesto:
Looks good and Travis likes it. Thanks!

PR #508, opened 2015-05-04
Fix seemingly obvious typo in the toLogLevel function

sylvanaar:
The logFlags for Verbose and Error seems to be transposed. Correcting on master, should probably be on the swift_support branch too


MaddTheSane:
Good catch.

rivera-ernesto:
Probably good for a 2.0.1 release soon?

Issue #361, opened 2014-10-10
Converted machThreadID instead of threadID


SergiyZhyrov:
In format: '"%s %s[%i:%s] %s", timestamp, appName, processID, threadID, logMsg'
Instead of threadID, logged machThreadID = pthread_mach_thread_np(pthread_self()); converted to unsigned hexadecimal.

To get threadID: 
__uint64_t threadid;
pthread_threadid_np(pthread_self(), &threadid);

NSLog use 'threadID'

Issue #311, opened 2014-07-20
Thread Id print mismatch between NSLog and DDTTYLogger


maximveksler:
I’m just began playing with CocoaLumberjack git master, running on Xcode 6 Beta 3, iOS8, iPhone 5s. It seems that Lumberjack and NSLog can’t agree on the mach port thread id.

    - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:    (NSDictionary *)launchOptions {
        
        [DDLog addLogger:[DDTTYLogger sharedInstance]];
        
        DDLogInfo(@"App loaded");
        NSLog(@"App loaded");
        
        return YES;
    }

This outputs:

    2014-07-20 02:32:17:085 Festival[931:807] App loaded
    2014-07-20 02:32:17.086 Festival[931:224090] App loaded

I would expect same value for both, probably the one NSLog prints, in this case `224090`

Not sure if this is a bug or a feature request, looking at the source code I do see that DDTTYLogger forces the machThreadID to be chopped at 8 characters.

ahknight:
Nope.  It's this (iOS 8 changed which thread ID it used in NSLog): http://stackoverflow.com/a/8995737/108928

rivera-ernesto:
Could you try to make a pull request?

bpoplauschi:
Fixed by #514

PR #514, opened 2015-05-19
Fix NSLog threadid mismatch with iOS 8+/OSX 10.10+

jeanregisser:
See the original issue https://github.com/CocoaLumberjack/CocoaLumberjack/issues/311


jeanregisser:
This would close #361 as well.

rivera-ernesto:
Looks, good. Thanks!

PR #515, opened 2015-05-19
Prevent duplicate log entries when both DDASLLogCapture and DDASLLogger are used

jeanregisser:
When both `DDASLLogCapture` and `DDASLLogger` are active and you have another logger (for instance the file or tty logger),
using a `DDLog(xxx)` statement would result in 2 log entries with the same message in other loggers.

Indeed, the first (expected) entry would be sent to the registered loggers and then the `DDASLLogger` would send it to the ASL facility, after that the `DDASLLogCapture` would pick it up and it would again be sent to the registered loggers (note that in that case `DDASLLogger` would rightfully skip it since it comes from the capture).

So here I've marked messages sent by `DDASLLogger` with a custom key that's used for filtering in `DDASLLogCapture`. Hence preventing the duplicate messages.

### DIFF
--- a/Classes/DDASLLogCapture.m
+++ b/Classes/DDASLLogCapture.m
@@ -95,6 +95,9 @@ static void (*dd_asl_release)(aslresponse obj);
     const char param[] = "7";  // ASL_LEVEL_DEBUG, which is everything. We'll rely on regular DDlog log level to filter
     
     asl_set_query(query, ASL_KEY_LEVEL, param, ASL_QUERY_OP_LESS_EQUAL | ASL_QUERY_OP_NUMERIC);
+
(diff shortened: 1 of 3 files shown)


rivera-ernesto:
Could you please rebase this? We just merged ASL capture code from #512 that may be conflicting.

jeanregisser:
Rebase done.

rivera-ernesto:
Still marked as conflicting below.

jeanregisser:
Oh indeed, cause more stuff has been merged since I did the rebase ;)

jeanregisser:
Should be good now!

rivera-ernesto:
Ok, merging this. Please give it a good try.

We will need some cleanup and fix unit tests before making a new release.

PR #518, opened 2015-05-22
Use an absolute path to identify the modulemap.

bjhomer:
In some cases when adding CocoaLumberjack as a submodule, Xcode can get confused and attempts to find the modulemap file in the wrong directory (e.g. `Mobile/Lumberjack/CocoaLumberjack.modulemap`, which does not exist). I'm not sure why happens or in exactly which situations, but using an absolute path (via `$PROJECT_DIR`) makes our intention more explicit and avoids this problem.


bjhomer:
Hm. It appears that some Travis CI tests failed, but I don't see how those failures could be related to the changes in this commit. Is the `master` branch passing its test?

rivera-ernesto:
Yes, they are nor related. Unit Tests are broken because of a race condition.

PR #520, opened 2015-05-24
Change code samples to use DDLogFlagWarning

Kumfert:
The code samples use DDLogFlagWarn, which isn't part of the DDLogFlag enum.
Switch to use DDLogFlagWarning (with the ING) instead. 
Also added spaces to keep things lining up pretty.


rivera-ernesto:
Thanks!

PR #521, opened 2015-05-29
Carthage support

MichaelMcGuire:
It looks like #446 has stalled and a request was made to fork and fix merge conflicts.  This pull request does that.  I was able to create a Cartfile and use my fork to build and add via Carthage.  I was also able to build and run the tests in the Lumberjack project, so I think it is good to go.


rivera-ernesto:
Thanks a lot!

PR #522, opened 2015-06-01
Fixes compilation issues on Swift 1.2

guidomb:
I had to add the 'context' and 'tag' parameters to DDLogXXX
functions to make it compile. It seems that this is a bug in
the Swift compiler with optional arguments and the use
of @autoclosure.

### DIFF
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -46,7 +46,7 @@ extension DDLogFlag {
 
 extension DDMultiFormatter {
     public var formatterArray: [DDLogFormatter] {
-        return self.formatters as [DDLogFormatter]
+        return self.formatters as! [DDLogFormatter]
     }
 }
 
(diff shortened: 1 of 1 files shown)


guidomb:
I have just realised that there is a 'swift-support' branch but I noticed that is not up-to-date with `master`. I don't know if tha branch adds anythings else but at least with this patch we can successfully compile CocoaLumberjack when using it as a dependency though Carthage.

rivera-ernesto:
@MaddTheSane ?

MaddTheSane:
This patch would make CocoaLumberjack on `master` compile on Xcode 6.3, but any Swift code using the Swift methods wouldn't log anything: The code in `SwiftLogMacro` is commented out.

It probably would be a good idea to merge the `swift_support` branch to `master`. Unless people are using an older version of Xcode.

guidomb:
@MaddTheSane @rivera-ernesto Ohh so sorry guys I forgot I commented that while I was fixing the compile error. I've just ammended the commit. Sorry.

rivera-ernesto:
We should definitely merge it. We can do that after #525.

rivera-ernesto:
@guidomb Please check if after #525 merge this should be updated/merged/closed.

bpoplauschi:
I've accidentally merged this as it was, into master. So I reverted it with a new commit (sorry about that). We need to remake this change on the swift_support branch. @guidomb could you do this?

guidomb:
I think that this can be closed if #525 is merged but the problem is that #525 is not building all the targets due to compilation errors.

rivera-ernesto:
Yes, swift_support was already merged. We should delete that branch after we add these "lost" commits (#531).

PR #526, opened 2015-06-04
improper prefix header path for mobile app file

idanhaviv:
the project didn't compile. i changed the prefix header path and file name to adapt to its actual name and location.


rivera-ernesto:
Thanks.

PR #530, opened 2015-06-11
Move DD_LEGACY_MACROS check to DDLegacyMacros.h

timbodeit:
When applying the new `use_frameworks!` option in Cocoapods, a bridging header
is automatically generated. This bridging header must import `DDLegacyMacros.h`
directly.
This leads to `CocoaLumberjack 1.9.x legacy macros enabled.` and `macro
redefined` warnings.

Therefore it is not enough to merely check the flag in DDLog.h.


rivera-ernesto:
Looks good. Thanks!

PR #536, opened 2015-06-20
Fix memory leaks in DDTTYLogger, add self annotations to blocks

madsolar8582:
Fixes some memory leaks in `DDTTYLogger` and add explicit `self` to blocks that implicitly retained it.

### DIFF
--- a/Classes/DDAbstractDatabaseLogger.m
+++ b/Classes/DDAbstractDatabaseLogger.m
@@ -217,7 +217,7 @@
 
     dispatch_sync(globalLoggingQueue, ^{
         dispatch_sync(self.loggerQueue, ^{
-            result = _saveThreshold;
+            result = self->_saveThreshold;
         });
     });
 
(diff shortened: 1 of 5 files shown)


rivera-ernesto:
Memory leak fixes look good, but not sure about how `self->` annotations are needed or could impact performance.
We had some [discussions](https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=properties+performance&type=Issues&utf8=✓) about using properties and `self.` all over the library too, and decided to keep direct access in places called frequently to have a better performance (also the original author's point of view).

madsolar8582:
`self->_ivar` is equivalent to `_ivar` as long as self hasn't been redefined in an inner scope (`self` is the implicit first function parameter). In the case of this code, `self` has not been redefined so the statements are equivalent. However, if you prefer the statements without the explicit `self` I can change them back.

rivera-ernesto:
Yes, probably switching back is better. For instance standard Objective-C setters don't use `self->` neither and would be hard to keep consistency with other contributors' code.

bpoplauschi:
Thanks @madsolar8582

PR #538, opened 2015-06-23
Add armv7s to static library

nrbrook:
I'm using Cocoa Lumberjack in a static framework distributed to other developers. I want to include armv7s in that framework in case developers want to use armv7s in their own Apps. However, the standard architectures in Xcode 6 and up don't include armv7s, and so the Cocoa Lumberjack targets don't compile for it.

For the dynamic framework the extra architecture consumes more space in the App bundle (750 KB vs 533 KB without), but for the static library I don't think there is any argument against including it as it will be stripped from the final App if not needed.

In this pull request I've just added armv7s to the static library target.

### DIFF
--- a/Lumberjack.xcodeproj/project.pbxproj
+++ b/Lumberjack.xcodeproj/project.pbxproj
@@ -1250,6 +1250,10 @@
 		18F3BFE91A81E06E00692297 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ARCHS = (
+					"$(ARCHS_STANDARD)",
+					armv7s,
+				);
 				CLANG_CXX_LANGUAGE_STANDARD = "gnu++0x";
 				CLANG_CXX_LIBRARY = "libc++";
 				GCC_PREPROCESSOR_DEFINITIONS = (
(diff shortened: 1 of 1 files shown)


rivera-ernesto:
Is this something that will get added by default with Xcode 7? If so you better switch to our `swift_2.0` branch instead.

nrbrook:
No, this is unrelated to Xcode 7.

rivera-ernesto:
Then why isn't `armv7s` part of `ARCHS_STANDARD`? I would prefer to keep default values as much as possible.

nrbrook:
From [Apple dev forums](https://devforums.apple.com/thread/244407?tstart=0)

> The reason for including armv7s would be marginal backward compatibility. But the general differences between armv7 and armv7s instruction sets are minor. So if you choose not to include armv7s, the targeted armv7 machine code still runs fine on 32 bit A6 devices, and hardly one will notice performance gap.

Developers have requested that I provide armv7s in the framework I produce, and to do that I also need Cocoa Lumberjack to compile armv7s. There is no reason not to, other than it not being the default value (but it includes the default values)

nrbrook:
If you don't want to merge, I will just do a fork for my own needs.

rivera-ernesto:
I see. Weird decision from Apple.
As you say including more slices in a library has nothing wrong at all.

nrbrook:
Think it's so App binaries are smaller so download quicker. For libraries it's a different matter. Cheers!

PR #545, opened 2015-06-25
Swift 2 fixes

MaddTheSane:
The following are fixes/enhancements to CocoaLumberjack for Swift 2.0/Xcode 7.

Notably, *DDLogMacros.h* can be set as a textual header, making defining custom macros for CocoaLumberjack, such as `ddLogLevel`, less troublesome.

### DIFF
--- a/Classes/CocoaLumberjack.modulemap
+++ b/Classes/CocoaLumberjack.modulemap
@@ -1,4 +1,0 @@
-framework module CocoaLumberjack {
-    header "CocoaLumberjack.h"
-    export *
-}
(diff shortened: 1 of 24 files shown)


rivera-ernesto:
Travis errors here I think are only due to Travis not supporting Swift 2.0 yet right?

MaddTheSane:
It looks like it. The Travis log doesn't mention the 10.11 SDK, which means it probably doesn't have Xcode 7 installed on it.

rivera-ernesto:
Yup. Let's merge this as it targets a non master branch. When Xcode 7 is out we can merge it back to master.

PR #551, opened 2015-07-06
Don't override defines in case they're already set at project level

Autoc0diq:


Classes/DDFileLogger.m
```diff
@@ -31,13 +31,15 @@
 // So we use primitive logging macros around NSLog.
 // We maintain the NS prefix on the macros to be explicit about the fact that we're using NSLog.
 
-#define LOG_LEVEL 2
+#ifndef DD_NSLOG_LEVEL
+    #define DD_NSLOG_LEVEL 2
+#endif
```
rivera-ernesto: I would prefer a:

```objc
#ifndef LOG_LEVEL
    #define LOG_LEVEL 2
#endif
```

instead.

Autoc0diq: LOG_LEVEL is very generic and it could be confused with the actual Cocoalumberjack log level. Here this define is specifically only used for NSLogs which is why I suggested renaming it to be clearer.

rivera-ernesto: Well, this is internal to a `.m` file so we can use whatever we want, right.

Autoc0diq: Well, the whole point of not redefining it if it's already defined is that it could be defined earlier, like let's say in Xcode build settings for a specific configuration (DEBUG). In this case, the naming is important to not be confusing.

PR #566, opened 2015-07-22
Flag unused parameters as being unused to silence strict warnings

hughescr:
Turning on strict warnings causes complaints about unused parameters.  Flag those parameters as unused to eliminate those warnings.


bpoplauschi:
Thanks @hughescr.

It looks like Travis is failing, but it's not related to this change. I see a lot of other projects failing on Travis builds due to: 
```
pod setup --silent

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

hughescr:
...always fun times when your build server takes >10 minutes to change directories :)

PR #568, opened 2015-07-23
typo in parameter's variable name fixed

krypt-lynx:
fixed misleading typo


rivera-ernesto:
I don't think this was a typo.

krypt-lynx:
This seems to be issue because parameter's counterpart is named "asynchronous" as like as another parameters in this methods group.
And it passed as parameter as variable named "asynchronous"

DD-P:
FWIW, I support this pull request, it looks as if the .h declaration doesn't match the .m definition.  (I don't think it would affect functionality, just the reader's understanding.)

PR #576, opened 2015-08-14
APPLICATION_EXTENSION_API_ONLY to YES for Extensions

sinoru:
This will make CocoaLumberjack can be linked with Application Extensions (Today, watchOS, etc) without this warning:
```
ld: warning: linking against dylib not safe for use in application extensions: {path_to_CocoaLumberjack_framework}
```


ksuther:
I'm not sure why the CI failed on this request, but this would be helpful to have merged.

sinoru:
Because master branch is already failed. Current image of Travis-CI's Xcode doesn't support Swift 1.2. Need to set `osx_image: beta-xcode6.3` on `.travis.yml`

Issue #383, opened 2014-10-30
Swift Compatibility Status


rivera-ernesto:
Following #299, #338, #358 and more broadly 2.x.x discussion in #225.

- [x] CocoaLumberjack.swift and CocoaLumberjackSwift framework wrappers (#362, #382).

Ideally users would install the Objective-C CocoaLumberjack library though CocoaPods, then either manually add CocoaLumberjack.swift, link CocoaLumberjackSwift?? Until there is official CocoaPods support for Swift.

To do:

- [x] Fix iOSSwift demo project.
- [x] Add Travis build steps for Swift code. iOS 8 SDK is already supported by Travis, OS X 10.10 SDK not yet.
- [x] Merge mobile and desktop frameworks/static libraries. iOS now supports frameworks as well so we could probably move to a single framework or at least single project.
- [x] Move framework/static libraries demo projects out to the Demos directory.

mackuba:
What version/branch/tag should I use at the moment to get Swift support?

MaddTheSane:
For Xcode 6.3 and Swift 1.2, the [swift_support](https://github.com/CocoaLumberjack/CocoaLumberjack/tree/swift_support) branch will compile.

guidomb:
I think that merging #466 will help using Cocoalumberjack in several Swift project / libraries.

rivera-ernesto:
I would like to but I was waiting for an answer from the contributor. If you want you can contribute to that pull request too so we can merge it.

PR #581, opened 2015-08-19
Add watchOSSwiftTest and some changes on Lumberjack.xcodeproj

sinoru:
I tried to add watchOSSwiftTest for testing on watch. And I realized that watchOS target requires another binary than that iOS's one. So we can't use same target for iOS application and watchOS application.

So, I create CocoaLumberjack-watchOS, CocoaLumberjackSwift-watchOS. After that, There are to many targets just for framework, So I merge mobile, desktop framework as mentioned on #383.

I didn't test watchOSSwiftTest on real device yet, but it works well on Watch Simulator.


rivera-ernesto:
Sounds good. Also this branch is ok for bigger changes.

rivera-ernesto:
@sinoru Could you take a look at #582?

sinoru:
@rivera-ernesto Oh, Merge CocoaLumberjack-iOS and CocoaLumberjack target can make some problem on Carthage...
I'll roll back that part.

Issue #53, opened 2012-08-21
semaphore_wait_trap : failed to resume in time


sgadhvi:
I get the below error while resuming the application there is some code in on resume which does the loging. but after showing splash screen app crashes with below logs.
What could be going wrong?

0   libsystem_kernel.dylib        	0x306eb060 semaphore_wait_trap + 8
1   libdispatch.dylib             	0x33879472 _dispatch_group_wait_slow + 146
2   libdispatch.dylib             	0x338793d2 dispatch_group_wait$VARIANT$mp + 22
3   XXXX                          	0x0012d38c +[DDLog lt_log:] (DDLog.m:959)
4   XXXX                          	0x0012c824 __+[DDLog queueLogMessage:asynchronously:]_block_invoke_1 (DDLog.m:476)
5   libdispatch.dylib             	0x338778c8 _dispatch_barrier_sync_f_invoke + 20
6   libdispatch.dylib             	0x33877728 dispatch_barrier_sync_f$VARIANT$mp + 56
7   libdispatch.dylib             	0x33877348 dispatch_sync_f$VARIANT$mp + 12
8   libdispatch.dylib             	0x338779f2 dispatch_sync$VARIANT$mp + 26
9    XXXX                          	0x0012c7d0 +[DDLog queueLogMessage:asynchronously:] (DDLog.m:484)
10  XXXX                         	0x0012ca8a +[DDLog log:level:flag:context:file:function:line:withTag:format:] (DDLog.m:554)

rivera-ernesto:
Can someone share a simple project that reproduces the bug?

bpoplauschi:
Closing, due to lack of response. Please comment and we can reopen it if you can provide a simple demo project.

Issue #33, opened 2012-05-03
Crash in DDLogFormatter subclass when accessing logMessage->timestamp


lkiss80:
I created a custom log formatter that is nearly identical to the example you posted under the wiki section.  It seemed to work well so I went ahead and released the app.  After some use, I am now experiencing an intermittent crash in my custom log formatter when access the timestamp instance variable of the logMessage passed into formatLogMessage.  Specific, this line here is causing the crash, and it's rooted within the timestamp.

```
NSString *dateAndTime = [dateFormatter stringFromDate:(logMessage->timestamp)];
```

I've only noticed the issue when opening the app, and it's coming out of the background state.  I haven't experienced it during normal use of the app once it's been running for a second or two.  For what it's worth, when the app first comes out of a background state, I communicate with a remote server to retrieve any updated config info.  It is typically a call to DDLogVerbose from within one of these API Interface functions.

Here is my complete custom formatter...

``` objectivec
@implementation iNotaryLogFormatter

- (id)init
{
    self = [super init];
    if (self) {
        dateFormatter = [[NSDateFormatter alloc] init];
        [dateFormatter setFormatterBehavior:NSDateFormatterBehavior10_4];
        [dateFormatter setDateFormat:@"yyyy-MM-dd HH:mm:ss:SSS"];
    }
    return self;
}

- (NSString *)formatLogMessage:(DDLogMessage *)logMessage
{
    NSString *logLevel;
    switch (logMessage->logFlag) {
        case LOG_FLAG_ERROR: logLevel = @"E"; break;
        case LOG_FLAG_WARN: logLevel = @"W"; break;
        case LOG_FLAG_INFO: logLevel = @"I"; break;
        default: logLevel = @"V"; break;
    }

    NSString *dateAndTime = [dateFormatter stringFromDate:(logMessage->timestamp)];
    NSString *logMsg = logMessage->logMsg;
    NSString *filename = [NSString stringWithUTF8String:logMessage->file];

    return [NSString stringWithFormat:@"%@ %@ | %@ [%d] %@\n", 
            logLevel, dateAndTime, 
            [filename lastPathComponent],
            logMessage->lineNumber,
            logMsg];
}

- (void)dealloc
{
    [dateFormatter release];
    [super dealloc];
}
```

PR #587, opened 2015-09-01
Surely we should be adding 1, not 0?

tshowalter:
The thread safety counter is being incremented by zero...

### DIFF
--- a/Classes/Extensions/DDDispatchQueueLogFormatter.m
+++ b/Classes/Extensions/DDDispatchQueueLogFormatter.m
@@ -94,7 +94,7 @@
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 - (NSString *)stringFromDate:(NSDate *)date {
-    int32_t loggerCount = OSAtomicAdd32(0, &_atomicLoggerCount);
+    int32_t loggerCount = OSAtomicAdd32(1, &_atomicLoggerCount);
 
     NSString *calendarIdentifier = nil;
 


rivera-ernesto:
Seems logic.

> `atomic add, incre-ment, increment,
     ment, decrement, or, and, xor, compare and swap, test and set, test and
     clear, spinlocks, and lockless queues`

> `OSAtomicAdd32(int32_t theAmount, volatile int32_t *theValue);`

Wondering why it was unnoticed and what was the result of this bug.

tshowalter:
Well, in our case the result was a random crash in NSDateFormatter, maybe one time in 30.

rivera-ernesto:
Let's give it a try. Could be related to #53.

erysaj:
Guys this change is totally wrong...
`_atomicLoggerCount` is counting (as variable name is implying) the loggers current formatter instance has been added to (hence atomic increments/decrements in corresponding delegate methods).
Now take a look at `stringFromDate:`. The method must deal with the fact that the NSDateFormatter is not thread safe (I've recently stumbled upon a related issue #33; I also bet that it is the crash @tshowalter experienced).
One of proposed (and allegedly working) solutions was to use `[NSThread threadDictionary]` to store date formatter instances per thread. Implementation of `DDDispatchQueueLogFormatter` tries to optimise overhead: if formatter is used by only 1 logger there is no need in thread-safety precautions, as logger processes log message on a serial dispatch queue.
To decide which approach to take you can't just read an ivar, it is modified by atomic functions so naturally you need a special way to read its value (with memory barrier). The simplest way seems to be just adding  `0` atomically (`OSAtomicAdd*` returns a result). Incrementing by `1`is a logical mistake: `stringFromDate:` is called on each log message and as a result `_atomicLoggerCount` will be nowhere close to the actual logger count.

Still this hybrid approach does not seem to work. Statistics for the app I develop shows crashes in `NSDateFormatter` and I can't figure out how exactly it happens considering all the taken measures
(we use a custom formatter that follows the same implementation pattern). My fix was to create
a separate formatter for each logger, though the fix is not in production yet, so I can't tell for sure that it helped. I think an acceptable solution would be to specify if thread-safety is required from formatter in its constructor and add assertion for the situation when users adds non-threadsafe instance to the 2nd logger.

P.S. while writing this comment I looked through reference issue and found [this wiki article](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/Documentation/CustomFormatters.md)

rivera-ernesto:
We really have a long-standing problem with this formatter so any contribution is greatly appreciated.

rivera-ernesto:
Related issues and pull requests: https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=NSDateFormatter&type=Issues&utf8=✓

erysaj:
> any contribution is greatly appreciated.

I've created #621 to show what I had in mind

bpoplauschi:
@erysaj Only now I totally understood your comments. Indeed, changing
`int32_t loggerCount = OSAtomicAdd32(0, &_atomicLoggerCount);` to `int32_t loggerCount = OSAtomicAdd32(1, &_atomicLoggerCount);` was a big mistake. That was only a pseudo-sync mechanism to make sure we read the correct value. But I'm not sure it's 100% safe. The big minus here was this was not properly documented.

Also, agreeing with the solution discussed on your PR. I will participate to that but I wanted this thread to show the actual mistake.

PR #590, opened 2015-09-10
Fix for the warning "Method override for the designated initializer of the superclass '-init' not found"

ksuther:
Fixes a new warning on Xcode 7 about designated initializers.

### DIFF
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -1122,6 +1122,10 @@ static int exception_count = 0;
     return [[self alloc] initWithFilePath:aFilePath];
 }
 
+- (instancetype)init {
+    @throw nil;
+}
+
 - (instancetype)initWithFilePath:(NSString *)aFilePath {
     if ((self = [super init])) {
         filePath = [aFilePath copy];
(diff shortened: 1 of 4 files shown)


rivera-ernesto:
I think this should be pull requested to the `swift_2.0` branch.

ksuther:
I can do that, but what's the reason? I'm running off of master and see the warning there.

Classes/DDFileLogger.m
```diff
@@ -1122,6 +1122,10 @@ + (instancetype)logFileWithPath:(NSString *)aFilePath {
     return [[self alloc] initWithFilePath:aFilePath];
 }
 
+- (instancetype)init {
+    @throw nil;
```
rivera-ernesto: Because this is not supported in swift 1.2.

ksuther:
It looks like the header part of this was already added to the swift_2.0 branch in commit 58ae948213e76a336473825256055b75d8da30b3. I'll close this. I assume the swift_2.0 branch will be getting merged into master relatively soon? I'm not actually using Swift at all, I just don't want these extraneous warnings appearing when I build :)

PR #597, opened 2015-09-18
Make build work on tvOS

ranterle:
tvOS will not support `host_info()` anymore. At the same time it will support `[[NSProcessInfo processInfo] activeProcessorCount]` so this can be used instead.

### DIFF
--- a/Classes/DDLog.m
+++ b/Classes/DDLog.m
@@ -140,15 +140,20 @@ static NSUInteger _numProcessors;
         // Figure out how many processors are available.
         // This may be used later for an optimization on uniprocessor machines.
 
+        NSUInteger one    = (NSUInteger)1;
+        NSUInteger result = one;
+#if TARGET_OS_TV
+        if ([NSProcessInfo class]) {
+            result = [[NSProcessInfo processInfo] activeProcessorCount];
+        }
+#else
         host_basic_info_data_t hostInfo;
         mach_msg_type_number_t infoCount;
 
         infoCount = HOST_BASIC_INFO_COUNT;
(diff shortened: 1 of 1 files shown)


rivera-ernesto:
`activeProcessorCount` has been available since iOS 2, so I wonder why we haven't used it ever since.

rivera-ernesto:
`activeProcessorCount` has been available since iOS 2, so I wonder why we haven't used it ever since.

rivera-ernesto:
`activeProcessorCount` has been available since iOS 2, so I wonder why we haven't used it ever since.

ranterle:
Right @rivera-ernesto - I had only recognized it in iOS 9 and didn't want to change too much except for tvOS devices. Updated to use `NSProcessInfo` on iOS as well.

rivera-ernesto:
Could you try to use the new code on iOS or OS X? A simple breakpoint in the only [place](https://github.com/ranterle/CocoaLumberjack/blob/master/Classes/DDLog.m#L668) the value is used would be enough to check if it works for all systems.

ranterle:
@rivera-ernesto yes it's called. Checked in a demo project on my iPhone 6 running iOS 9.

rivera-ernesto:
In that case let's just replace the old code with the new calls so we don't need AppleTV-specidifc code.

ranterle:
Ok, removed compiler macros altogether, however Mac OS before 10.5 does not support `[NSProcessInfo processorCount]` so protected against that.

ranterle:
@rivera-ernesto created, see #604

PR #604, opened 2015-10-01
Remove redundant check for processorCount availability

ranterle:
As pointed out in #597 the check whether `processorCount` is available is not necessary anymore since min version for OS X is 10.7 and all other platforms supported it from the start.

### DIFF
--- a/Classes/DDLog.m
+++ b/Classes/DDLog.m
@@ -141,12 +141,7 @@ static NSUInteger _numProcessors;
         // This may be used later for an optimization on uniprocessor machines.
         
         NSUInteger one    = (NSUInteger)1;
-        NSUInteger result = one;
-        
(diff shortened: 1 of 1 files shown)


Classes/DDLog.m
```diff
@@ -141,12 +141,7 @@ + (void)initialize {
...
         
         NSUInteger one    = (NSUInteger)1;
-        NSUInteger result = one;
-        
-        NSProcessInfo *processInfo = [NSProcessInfo processInfo];
-        if ([processInfo respondsToSelector:@selector(processorCount)]) {
-            result = processInfo.processorCount;
-        }
+        NSUInteger result = [NSProcessInfo processInfo].processorCount;
         _numProcessors = MAX(result, one);
```
rivera-ernesto: We could also use a `1` here directly.

PR #605, opened 2015-10-03
Add tvOS targets on project

sinoru:
This will allow include tvOS frameworks on projects, and let Carthage build tvOS framework.

### DIFF
--- a/Classes/DDLog.m
+++ b/Classes/DDLog.m
@@ -144,15 +144,20 @@ static NSUInteger _numProcessors;
         // Figure out how many processors are available.
         // This may be used later for an optimization on uniprocessor machines.
 
+        NSUInteger one    = (NSUInteger)1;
+        NSUInteger result = one;
+#if TARGET_OS_TV
+        if ([NSProcessInfo class]) {
+            result = [[NSProcessInfo processInfo] activeProcessorCount];
+        }
+#else
         host_basic_info_data_t hostInfo;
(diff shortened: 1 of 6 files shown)


Classes/DDLog.m
```diff
@@ -875,6 +866,13 @@ @implementation DDLogMessage
 #define USE_DISPATCH_CURRENT_QUEUE_LABEL YES
 #define USE_DISPATCH_GET_CURRENT_QUEUE   YES
 
+#elif TARGET_OS_TV
```
rivera-ernesto: Maybe it would be cool better to merge these conditions as `#if TARGET_OS_IOS || TARGET_OS_WATCH`.

Classes/Extensions/DDMultiFormatter.m
```diff
@@ -26,6 +26,9 @@
 #elif TARGET_OS_WATCH
 // Compiling for watchOS
 #define NEEDS_DISPATCH_RETAIN_RELEASE 0
+#elif TARGET_OS_TV
+// Compiling for tvOS
+#define NEEDS_DISPATCH_RETAIN_RELEASE 0
```
rivera-ernesto: Same here.

sinoru:
You mean like this?

rivera-ernesto:
Yes, specially since both platforms share the same behavior.

rivera-ernesto:
We may need to create a separate branch for this one following #609.

PR #613, opened 2015-10-09
Kill unused function warnings from DDTTYLogger.h

kdubb:
Use "inline" qualifier on DDMakeColor to get rid of unused function warnings


rivera-ernesto:
Nice.

Issue #592, opened 2015-09-15
Designated Initializer Issues


stephencelis:
Both DDLogMessage and DDLogFileInfo define designated initializers (using `NS_DESIGNATED_INITIALIZER`), but do not fulfill the requirement that their superclass's initializer (`init`) is overridden. This causes two of these warnings:

> Method override for the designated initializer of the superclass '-init' not found

Any reason not to drop `NS_DESIGNATED_INITIALIZER` in these two cases? Designated initializers seem to be most important for classes intended to be subclassed, and I'm not sure these are.

rivera-ernesto:
You are right. Do you want to propose a pull request?

stephencelis:
Done!

timbodeit:
When trying to merge the current `swift_2.0` branch into `master`, 58ae9482 conflicts with b765caf.

To me it looks like you both meant to fix the same issue in two different ways.
You dropped the `NS_DESIGNATED_INITIALIZER` argument, whereas @sinoru added
`- (instancetype)init NS_UNAVAILABLE;`

On first sight, I personally like @sinoru s approach better. 

Any input on why one is objectively better than the other or  how the merge conflict should be resolved?

stephencelis:
I also like @sinoru's approach. Hadn't noticed `NS_UNAVAILABLE` pop up in iOS 8.

rivera-ernesto:
One could merge `swift_2.0` into `master` in a fork, fix que conflict and pull request that to update `swift_2.0` here.

bpoplauschi:
@rivera-ernesto we need to take care of this fast, no point in delaying it. I will allocate some time for this (merge the 2.0 swift branch into master and cut a beta release).

rivera-ernesto:
I think we should be less afraid of releasing both a 2.0.2 and a 2.1.0 beta. Thanks @bpoplauschi !

timbodeit:
I'm currently working on getting the merge conflicts resolved. Will push a proposed resolution in around 20 minutes.

timbodeit:
@bpoplauschi See [#614](https://github.com/CocoaLumberjack/CocoaLumberjack/pull/614)

PR #619, opened 2015-10-13
Re-add NS_DESIGNATED_INITIALIZER

sinoru:
I found that one NS_DESIGNATED_INITIALIZER removed, so re-added it.


rivera-ernesto:
There's [another approach](https://github.com/CocoaLumberjack/CocoaLumberjack/issues/592#issuecomment-146835164)  by @sinoru that involves using `- (instancetype)init NS_UNAVAILABLE`. We should choose one or the other and stuck with it.

bpoplauschi:
I missed that. Too many changes I guess :)

Issue #452, opened 2015-02-10
setFormattedBehavior: call not needed in custom formatter doc


jlnbuiles:
i could be wrong, but after looking at apple's documentation (https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSDateFormatter_Class/index.html#//apple_ref/occ/clm/NSDateFormatter/defaultFormatterBehavior) i believe this line isn't needed here, given that the default formatted behavior is the same as what you are setting it to be.

"For iOS and for OS X applications linked against OS X v10.5 and later, the default is NSDateFormatterBehavior10_4."

rivera-ernesto:
You're probably right. Would you like to make a pull request ;)

jlnbuiles:
done! https://github.com/CocoaLumberjack/CocoaLumberjack/pull/454

Issue #44, opened 2012-06-25
EXC_BAD_ACCESS because NSDateFormatter is not thread-safe in DispatchQueueLogFormatter


GrahamDennis:
I've been using Cocoa Lumberjack with the `DispatchQueueLogFormatter` for a while, and just hit a crash which appears to be due to using a single `NSDateFormatter` simultaneously across two threads.

The relevant backtraces are:

The thread that crashed:

```
* thread #15: tid = 0x2f03, 0x01761f8a CoreFoundation`CFDateFormatterCreateStringWithAbsoluteTime + 74
    frame #0: 0x01761f8a CoreFoundation`CFDateFormatterCreateStringWithAbsoluteTime + 74
    frame #1: 0x01761f2c CoreFoundation`CFDateFormatterCreateStringWithDate + 76
    frame #2: 0x00e6d888 Foundation`-[NSDateFormatter stringForObjectValue:] + 139
    frame #3: 0x00e6d7f4 Foundation`-[NSDateFormatter stringFromDate:] + 44
    frame #4: 0x0007d89f BibDeskReader`-[DispatchQueueLogFormatter formatLogMessage:] + 95 at DispatchQueueLogFormatter.m:183
    frame #5: 0x0007ca10 BibDeskReader`-[DDTTYLogger logMessage:] + 208 at DDTTYLogger.m:99
    frame #6: 0x0007a650 BibDeskReader`__16+[DDLog lt_log:]_block_invoke_0 + 80 at DDLog.m:615
    frame #7: 0x01e3c330 libdispatch.dylib`_dispatch_call_block_and_release + 15
    frame #8: 0x01e3df0c libdispatch.dylib`_dispatch_queue_drain + 460
    frame #9: 0x01e3dcb4 libdispatch.dylib`_dispatch_queue_invoke + 50
    frame #10: 0x01e3d402 libdispatch.dylib`_dispatch_worker_thread2 + 247
    frame #11: 0x91db0b24 libsystem_c.dylib`_pthread_wqthread + 346
```

The other thread that was using the same NSDateFormatter:

```
* thread #14: tid = 0x3003, 0x91dadaf6 libsystem_c.dylib`__mtx_droplock + 773
    frame #0: 0x91dadaf6 libsystem_c.dylib`__mtx_droplock + 773
    frame #1: 0x91dadc4c libsystem_c.dylib`pthread_mutex_unlock + 320
[...]

Issue #39, opened 2012-06-02
Date/Time Formatting Issue


neilt:
The problem is formatting of the date time functions. The following examples are from the console in Xcode using iPhone 5.1 simulator.  The log is also written to a file and it has the same anomaly sometimes, but not always at the same places.  These happen all the time in the logs.  Its not hard to find and provide examples.  This is using a fresh pull as of June 2, 09:41 Central time.  Xcode is Version 4.3.2 (4E2002).

For example from file log.
`
2012/06/02 09:23:29:222 [DCADatabasesViewController getDatabases]:256, Sort by displaySequence
2012/06/00 00:00:00:000 [DCADatabasesViewController tableView:numberOfRowsInSection:]:277, count of db's = 1
2012/06/155 09:23:29:288 [DCADatabasesViewController tableView:cellForRowAtIndexPath:]:284, -
`
Note the time is just wrong. Now the same info from Xcode. A lot of the time the errors are the same.
`
2012/06/02 09:23:29:222 [DCADatabasesViewController getDatabases]:256, Sort by displaySequence
2012/06/00 00:00:00:000 [DCADatabasesViewController tableView:numberOfRowsInSection:]:277, count of db's = 1
2012/06/155 09:23:29:288 [DCADatabasesViewController tableView:cellForRowAtIndexPath:]:284, -
`

Another example from Xcode.
`
2012/06/02 09:23:23:022 [DCADatabasesEditNameViewController viewDidAppear:]:70, -
2012/06/02 09:0023:29:171 [DCADatabasesEditNameViewController saveButtonAction]:136, -
2012/06/03 09:23:29:222 [DCADatabasesEditNameViewController viewWillDisappear:]:76, -
`
But in this case the following from the file log is ok.
`
2012/06/02 09:23:23:022 [DCADatabasesEditNameViewController viewDidAppear:]:70, -
2012/06/02 09:23:29:171 [DCADatabasesEditNameViewController saveButtonAction]:136, -
2012/06/03 09:23:29:222 [DCADatabasesEditNameViewController viewWillDisappear:]:76, -
`

I am using a custom formatter (essentially directly from the example code.)

```
- (id)init
{
    if((self = [super init]))
    {
        dateFormatter = [[NSDateFormatter alloc] init];
        [dateFormatter setDateFormat:@"yyyy/MM/dd HH:mm:ss:SSS"];
    }
    return self;
}

[...]

PR #621, opened 2015-10-13
Fix NSDateFormatter issues

erysaj:
Attempt to solve crashes in NSDateFormatter still observed in the wild.
Stop using so-called "hybrid approach" when implementation tries to determine in runtime what NSDateFormatter instance to use: the shared one or retrieved from thread dictionary. Now a library user must decide if thread-safe but more expensive behaviour is desired (default) or simple storing date formatter in ivar is sufficient (suitable for using formatter with a single logger).

A few more minor tweaks:
- make simple customisations (like changing date format) easy by exposing a method to configure used NSDateFormatter instance
- set date formatter's calendar once instead of doing that each time a log message is formatted
- use POSIX locale to have consistent timestamp. For example, currently on devices with Arabian regional settings you'll see digits that you've probably never seen before.

### DIFF
--- a/Classes/Extensions/DDDispatchQueueLogFormatter.h
+++ b/Classes/Extensions/DDDispatchQueueLogFormatter.h
@@ -24,6 +24,11 @@
 #import "DDLog.h"
 
 
+typedef NS_OPTIONS(NSUInteger, DDDispatchQueueLogFormatterOptions) {
+    DDDispatchQueueLogFormatterShareable = 1 << 0,
+};
+
+
 /**
  * This class provides a log formatter that prints the dispatch_queue label instead of the mach_thread_id.
  *
@@ -68,6 +73,8 @@
  * Configure using properties as desired.
  **/
 - (instancetype)init NS_DESIGNATED_INITIALIZER;
+
+- (instancetype)initWithOptions:(DDDispatchQueueLogFormatterOptions)options;
 
 /**
  * The minQueueLength restricts the minimum size of the [detail box].
--- a/Classes/Extensions/DDDispatchQueueLogFormatter.h
+++ b/Classes/Extensions/DDDispatchQueueLogFormatter.h
@@ -135,6 +135,7 @@ typedef NS_OPTIONS(NSUInteger, DDDispatchQueueLogFormatterOptions) {
  **/
 @interface DDDispatchQueueLogFormatter (OverridableMethods)
 
+- (void)configureDateFormatter:(NSDateFormatter *)dateFormatter;
 - (NSString *)stringFromDate:(NSDate *)date;
 - (NSString *)queueThreadLabelForLogMessage:(DDLogMessage *)logMessage;
 - (NSString *)formatLogMessage:(DDLogMessage *)logMessage;
--- a/Classes/Extensions/DDDispatchQueueLogFormatter.h
+++ b/Classes/Extensions/DDDispatchQueueLogFormatter.h
@@ -24,8 +24,9 @@
 #import "DDLog.h"
 
 
-typedef NS_OPTIONS(NSUInteger, DDDispatchQueueLogFormatterOptions) {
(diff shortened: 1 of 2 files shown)


Classes/Extensions/DDDispatchQueueLogFormatter.m
```diff
@@ -41,7 +43,22 @@ @implementation DDDispatchQueueLogFormatter
...
     if ((self = [super init])) {
-        _dateFormatString = @"yyyy-MM-dd HH:mm:ss:SSS";
+        _options = DDDispatchQueueLogFormatterShareable;
+
+        // We need to carefully pick the name for storing in thread dictionary to not
+        // use a formatter configured by subclass and avoid surprises.
+        Class cls = [self class];
+        SEL configMethodName = @selector(configureDateFormatter:);
+        Method configMethod = class_getInstanceMethod(cls, configMethodName);
+        for (;;) {
```
rivera-ernesto: Probably `while (YES)` may be better. Nor sure if `do` can be used.

Classes/Extensions/DDDispatchQueueLogFormatter.h
```diff
@@ -24,6 +24,11 @@
 #import "DDLog.h"
 
 
+typedef NS_OPTIONS(NSUInteger, DDDispatchQueueLogFormatterOptions) {
+    DDDispatchQueueLogFormatterShareable = 1 << 0,
+};
```
rivera-ernesto: Shouldn't this have at least two options?

Classes/Extensions/DDDispatchQueueLogFormatter.m
```diff
@@ -93,48 +117,54 @@ - (void)setReplacementString:(NSString *)shortLabel forQueueLabel:(NSString *)lo
 #pragma mark DDLogFormatter
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-- (NSString *)stringFromDate:(NSDate *)date {
-    int32_t loggerCount = OSAtomicAdd32(1, &_atomicLoggerCount);
+- (NSDateFormatter *)createDateFormatter {
+    NSDateFormatter *formatter = [[NSDateFormatter alloc] init];
+    [self configureDateFormatter:formatter];
+    return formatter;
+}
```
rivera-ernesto: Just to be sure, we may need several instances of `formatter`? Asking just in case we could have a singleton.

rivera-ernesto:
Related to #33, #53, #452, #296, #44, #39.

Classes/Extensions/DDDispatchQueueLogFormatter.m
erysaj: I'll rewrite using `while` with non-trivial condition and without `break`-s

Classes/Extensions/DDDispatchQueueLogFormatter.h
erysaj: It is a bit flag, there is 2 options already: flag ON or OFF
Though on the second though I think it'd be more intuitive if default behaviour corresponds to `0` options mask.

Classes/Extensions/DDDispatchQueueLogFormatter.m
erysaj: When created in "shareable" mode (meaning usable with multiple loggers) we'll have a dedicated NSDateFormatter instance per each thread where log formatter is used;
so yes, we may need several instances

Classes/Extensions/DDDispatchQueueLogFormatter.h
rivera-ernesto: I think we should have named options for both on and off. Usually the one that has a raw value of 0 will be default.

bpoplauschi:
@erysaj I think your implementation looks strong, so I will go ahead and merge it. We will keep an eye on this.

PR #627, opened 2015-10-21
Fix Issue #488: Support DDLog without AppKit Dependency (#define DD_CLI)

### LINKED ISSUE #488: DDTTYLoggeer's AppKit detection doesn't work & Path to CLIColor.h include wrong

bpoplauschi:
This PR replaces #497 which has conflicts. The author of the commit is still @weissi.

Fixes #488

### DIFF
--- a/Classes/DDTTYLogger.h
+++ b/Classes/DDTTYLogger.h
@@ -44,15 +44,15 @@
     #import <UIKit/UIColor.h>
     typedef UIColor DDColor;
     static inline DDColor* DDMakeColor(CGFloat r, CGFloat g, CGFloat b) {return [DDColor colorWithRed:(r/255.0f) green:(g/255.0f) blue:(b/255.0f) alpha:1.0f];}
-#elif __has_include(<AppKit/NSColor.h>)
-    // OS X with AppKit
-    #import <AppKit/NSColor.h>
-    typedef NSColor DDColor;
-    static inline DDColor* DDMakeColor(CGFloat r, CGFloat g, CGFloat b) {return [DDColor colorWithCalibratedRed:(r/255.0f) green:(g/255.0f) blue:(b/255.0f) alpha:1.0f];}
-#else
+#elif defined(DD_CLI) || !__has_include(<AppKit/NSColor.h>)
     // OS X CLI
     #import "CLIColor.h"
     typedef CLIColor DDColor;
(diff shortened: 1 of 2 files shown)


rivera-ernesto:
We should probably just leave 2.0.x separate from master and merge swift 2.0 to master and work from there from now on.

bpoplauschi:
Swift 2.0 is merged into master already :). I just left the branch there for people that were referring it explicitly. I got it merged right after 2.0.2 and before the 2.1.0 beta.

rivera-ernesto:
Right. We should probably delete the branch as well so people test the master instead.

bpoplauschi:
Ok.

PR #634, opened 2015-10-22
Merge tvOS branch

rivera-ernesto:
I think people upgrading from Xcode 6.x should jump to 7.1 now that it has been released.


bpoplauschi:
Xcode 7.1 was just released and the adoption will take some time. I don't think we should force our users into using the latest version of Xcode. I would postpone merging this branch into master for a while, especially until we release an official 2.1.0.

rivera-ernesto:
Sure, we can keep this open in the mean time.

bpoplauschi:
I've merged the branch into master, as we have released 2.1.0 and the next release 2.2.0 should probably include Xcode 7.1 and tvOS support. For backwards compatibility, I've created an [Xcode_7.0 branch](https://github.com/CocoaLumberjack/CocoaLumberjack/tree/Xcode_7.0)

bpoplauschi:
@rivera-ernesto do you see any potential issues in releasing a 2.2.0 from the current master (for the TVOS support)? I'm thinking that some people that didn't upgrade to Xcode 7.1 yet will not be able to build the library/framework and therefore it will break compatibility. In this light, we might want to change the version to 3.0.0. But then again, this was the same thing when we released 2.1.0 and we broke compatibility with Xcode 6.

My issue here is the semantic version and how people with 2.0.x will receive those updates, if those will cause issues for them.

bpoplauschi:
An interesting piece of info: https://gist.github.com/orta/1c607a2c46545244ab16, seems like Lumberjack is the 9th most used Cocoa library based on the number of apps that include it.

rivera-ernesto:
Good idea.

I think that in 2.x releases it's ok to ask to upgrade Xcode, change some API's. etc. So yes, we could release 2.2.0 and if someone wants to contribute to the 2.1.x or 2.0.x versions we can create branches for their pull requests if needed.

bpoplauschi:
Ok then, a 2.2.0 release we should make (Yoda phrasing).

PR #636, opened 2015-10-22
Enable bitcode

ichibod:
Enabling bitcode for compiling via carthage


Lumberjack.xcodeproj/project.pbxproj
```diff
@@ -1989,6 +1992,7 @@
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/Frameworks";
 				MODULEMAP_FILE = Framework/Lumberjack/CocoaLumberjack.modulemap;
+				ONLY_ACTIVE_ARCH = NO;
```
rivera-ernesto: This should be `YES` for Debug builds.

ichibod: If you're using carthage, which is one of the recommended install methods, shouldn't this be `NO`, so it can be run on all architectures as it's not built on the fly.

rivera-ernesto: But then why would you distribute a Debug, non-optimized version of the code?

ichibod: AFAIK, Carthage pre-compiles your frameworks, so if the framework is precompiled in debug for 32-bit, probably won't be optimal for the 64-bit variation - meaning the active arch would be whatever machine compiled it on? Carthage is new to me, so I could be misspeaking.

rivera-ernesto:
What I don't see is a reason to distribute compiled inefficient code.
You can distribute the Release build for all architectures and still use that with both Debug and Release configurations of your app as long as the required architectures are included. Which Release does and Debug by default doesn't because you're supposed to use it only for local builds.

Also people using the project directly will see Xcode suggesting to turn it back to `YES` or get slower compiling times when developing.

ichibod:
It's really a moot point because the carthage system won't work for my project.

But as far as my understanding of it is concerned, is carthage pre-compiles the framework (i.e. CL) and then you include those in your project. So you could have inefficient/inaccurate architectures compiled into your debug builds. And since this is simply for the pre-compiling before inclusion in a bigger project, the bigger project would have the flag set to yes, while this one would pre-compile for all (NO).

PR #642, opened 2015-10-27
Remove (escaping) from the Swift @autoclosure parameters

kballard:
This was added in a993d7cbb in a misguided attempt to avoid having to
say `logText()` when passing the parameter to `SwiftLogMacro` (see [1]).
Besides being unnecessary, `@autoclosure(escaping)` actually carries a
performance penalty, it breaks the ability to reference properties and
methods on `self` without an explicit `self.` (in string
interpolations), and it makes the function look, to its callers, as
though it actually evaluates the message outside of the scope of the
function, which would be a really terrifying thing if it actually did
that.

[1]: [GitHub comment](https://github.com/CocoaLumberjack/CocoaLumberjack/pull/492#issuecomment-91104180)

### DIFF
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -54,7 +54,7 @@ public func resetDefaultDebugLevel() {
     defaultDebugLevel = DDLogLevel.Verbose
 }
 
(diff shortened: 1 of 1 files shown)


bpoplauschi:
Thanks @kballard

PR #646, opened 2015-11-01
Updating documentation for manual installation

DD-P:
I've had a go at updating the Getting Started documentation - and in particular the Manual Installation instructions.

You'll see that I've moved around some of the later instructions for project integration (now called accessing and configuring CLJ).  I've also preferred the non key-based HTTPS URL schemes for the git clone and submodule examples to avoid the risk of permissions errors when getting the source.

I'd be pleased if the team could review these changes for accuracy and any oversights.  As our friends at Apple might say: Focus Areas are CocoaPods, Carthage and Swift ;)

### DIFF
--- a/Documentation/GettingStarted.md
+++ b/Documentation/GettingStarted.md
@@ -1,41 +1,81 @@
-_Getting started with the lumberjack framework._
+_Getting started with the CocoaLumberjack framework._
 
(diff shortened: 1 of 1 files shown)


bpoplauschi:
Thansk @DD-P

Issue #542, opened 2015-06-25
2.0.1


plandem:
again breaking backward compatibility?

Code that was working fine with 2.0.0, does not work at 2.0.1 anymore.
```
Error:(74, 3) use of undeclared identifier 'ddLogLevel'
```

bpoplauschi:
We had an issue with `ddLogLevel` which had to be fixed. Basically, you still need to define it.

plandem:
as i said....at 2.0.0 it's working fine. and yes, at my code i have:
```
#if DEBUG
	static const DDLogLevel ddLogLevel = DDLogLevelVerbose;
#else
	static const DDLogLevel ddLogLevel = DDLogLevelWarning;
#endif
```

bpoplauschi:
Where is your definition of `ddLogLevel`? I usually set it in the precompiled header, next to `#import <CocoaLumberjack.h>`

plandem:
it's long story. right now it's a library from another cocoapod. but, as i said, at 2.0.0 it's working fine.

markcleonard:
Have the same situation as @plandem. We use CocoaLumberjack in a shared framework between apps, which was working fine in 2.0.0 and is now broken in 2.0.1 for the same reason.

colinbrash:
We are experiencing the same issue.  I don't fully understand, "you still need to define [ddLogLevel]".  We did not have to define it in 2.0.0.  Changing this behavior in a patch version seems wrong.

bpoplauschi:
Guys, the fix from 2.0.0 that allowed you to skip defining `ddLogLevel` was reported as a bug https://github.com/CocoaLumberjack/CocoaLumberjack/issues/478. Since we didn't have a good fix for that, I thought we just revert it. Sorry for the inconvenience, but it should be an easy fix to make sure `ddLogLevel` is defined.

plandem:
but in my case i HAVE ddLogLevel

jtdavid:
I also have ddLogLevel and still have the issue as @plandem refered. I do have a private repo with CocoaLumberjack as dependency in my embed framework podspec...

bpoplauschi:
Guys, are you perhaps using Swift? I cannot understand why this issue occurs, if ddLogLevel is defined.

tbaranes:
Same issue here using Lumberjack in an Objective-C project. I didn't have it using the 2.0.0, and ddLogLevel is well defined...

markcleonard:
[...]

PR #647, opened 2015-11-03
Get rid of separate CocoaLumberjackSwift.framework for most platforms

kballard:
For iOS, watchOS, and tvOS, we can support Swift directly with CocoaLumberjack.framework instead of having a separate CocoaLumberjackSwift.framework. OS X still requires CocoaLumberjackSwift.framework because it has a deployment target of 10.7 but Swift requires 10.9 or later.


bpoplauschi:
I think those changes might break compatibility with the existing installs, so after our recent releases have also done this, not sure it this is wanted.
@rivera-ernesto what do you think?

rivera-ernesto:
I think that anything that simplifies the project is good. Let's prepare a 2.3.0 release allowed to break some small things like this and a default `ddLogLevel` (#648) in order to improve the library.

rivera-ernesto:
As for this pull request in particular, I think it should be broken into two different pull requests as modifying the Swift "macros" is not what this pull request describes.

Let's just keep the merge frameworks commits here.

DD-P:
If there are backward-incompatible changes (e.g. to the .xcodeproj), would a 3.0.0 version be preferred to maintain semantic versioning?

SemVer was being advocated in the past (#542).

kballard:
The macro changes actually are backwards-compatible. This is why I left `SwiftLogMacro` intact and just marked it as deprecated. The project changes are the backwards-incompatible part.

I almost left the `CocoaLumberjackSwift` modules intact as stubs, to preserve backwards compatibility, but I took them out entirely because I figured that was just too confusing. CocoaLumberjack already doesn't do a good job of explaining Swift support, so I figured it was better to kill the CocoaLumberjackSwift framework entirely so people wouldn't unnecessarily import it. It also speeds up the build a bit (because building those stub frameworks does take time).

rivera-ernesto:
@DD-P only v.v.X versions really need to be minor. We're not rewriting most classes as we did for 2.0.0 and we don't want to be in version 18 in a year, so I think it's ok to make a 2.3.0 release that requires some small migration and only for users that installed manually.

@kballard The macro changes look good, it is just that I think they should be made in a separate pull request to make it cleaner/easier to review/discuss/document.

kballard:
Ok I'll split it out.

kballard:
PR updated, #649 filed for the Swift cleanup.

rivera-ernesto:
Thanks!

PR #649, opened 2015-11-05
Clean up CocoaLumberjack.swift a little

kballard:
Rename SwiftLogMacro to _DDLogMessage instead, because SwiftLogMacro is
a weird name for this. Leave the old name behind but marked as
deprecated for now; it should be removed later. Give _DDLogMessage a
more sane argument ordering as well (and no default arguments; it's not
meant for end users to call).

Stop using string interpolation for the `StaticString`s. The comments
say it avoids an integer overflow warning, but I see no such warning
(and never have).

Rewrite `CurrentFileName()` in pure Swift instead of unnecessarily
bridging to Cocoa.

### DIFF
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -54,38 +54,49 @@ public func resetDefaultDebugLevel() {
     defaultDebugLevel = DDLogLevel.Verbose
 }
 
+@available(*, deprecated, message="Use one of the DDLog*() functions if appropriate or call _DDLogMessage()")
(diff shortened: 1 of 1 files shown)


kballard:
Note: This should be backwards-compatible, with any current users of `SwiftLogMacro` seeing a deprecation warning upon upgrade.

rivera-ernesto:
Could you take handle this pull request @MaddTheSane ?

Also not sure is its worth keeping backwards compatibility if the code is going to be cleaner and its trivial to update for users.

rivera-ernesto:
Sorry @kballard, seems like nobody reviewed this. Could you update the pull request?

kballard:
@rivera-ernesto PR updated.

rivera-ernesto:
Thanks, looking good.

johndpope:
Is anyone using swift the fantastic  nbulog https://github.com/CyberAgent/NBULog/issues/6 ?

rivera-ernesto:
I think I should rewrite NBULog in Swift. There's not much code to it.

PR #650, opened 2015-11-05
Add line showing how to add a custom formatter to a logger

dxclancy:
This would have been useful to me when reading this tutorial, so I thought i'd propose adding it!

Thanks!


rivera-ernesto:
Thanks!

PR #655, opened 2015-11-14
make the method names correspond to the one in source(DDLog.h)

sushichop:
"registered dynamic logging" is very helpful for me. But as you know, the document seems to be old. So I have updated it(I thought I'd like to contribute!).

Thank you.


bpoplauschi:
Thanks @sushichop

PR #657, opened 2015-11-17
Simplify usage and integration of the static library target

iabudiab:
IMHO, the configuration of the static library target is suboptimal and doesn't provide a sane default. 

This PR makes the following changes for the static library target:

- The product name is changed to `CocoaLumberjack` instead of the default which resolves to `CocoaLumberjack-iOS-Static`
- A copy-headers phase is added, which copies all the headers to `$(BUILD_PRODUCTS_DIR)/include/CocoaLumberjack`.
- This was also documented in the `Getting Started`

This simplifies the usage of the library in nested/complex project configurations, since the headers can always be found and correctly imported via the angle brackets, i.e.

`#import <CocoaLumberjack/CocoaLumberjack.h>`

provided that the `Header Search Paths` has the entry `$(BUILD_PRODUCTS_DIR)/include`, which many projects already use.

### DIFF
--- a/Documentation/GettingStarted.md
+++ b/Documentation/GettingStarted.md
@@ -48,9 +48,24 @@ _Please note, installation via CocoaPods or Carthage is much simpler and recomme
 	* Drag in the CocoaLumberjack.framework from the Lumberjack.xcodeproj products group
 	* _Note: be careful to include only your relevant platform product_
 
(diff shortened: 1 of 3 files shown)


rivera-ernesto:
Nice. Thanks!

Issue #658, opened 2015-11-18
CocoaLumberjack iOS framework compiles swift file


nrbrook:
The normal CocoaLumberjack framework compiles the swift file. This is a problem if the App project is not swift and does not have `EMBEDDED_CONTENT_CONTAINS_SWIFT` enabled.

I'm not sure if there is a reason the swift file is being compiled in the normal framework, but removing it seems ok in both swift and non-swift projects.

rivera-ernesto:
Could you make a pull request?

nrbrook:
Sorry, I thought I had. Done: https://github.com/CocoaLumberjack/CocoaLumberjack/pull/659

rivera-ernesto:
Closing this to move discussion to #659.

PR #659, opened 2015-11-19
Remove swift file from normal iOS framework

nrbrook:
Fixes https://github.com/CocoaLumberjack/CocoaLumberjack/issues/658


sinoru:
This will broke other swift project. How about revert #647

nrbrook:
This does need resolving as otherwise non-swift projects require the swift dylib which adds to the bundle size. There could perhaps be two separate framework targets.

rivera-ernesto:
I think that two targets makes sense as less projects stay Objective-C-only.

sinoru:
So revert #647 (db6a623) that merges CocoaLumberjack, CocoaLumberjackSwift can solve the issue. I think.

nrbrook:
I've just added a new commit which adds a separate swift framework target. I've tested on both swift and non-swift projects. There is no test target for the non-swift framework though.

rivera-ernesto:
The advantage of (partially) reverting #647 would be ensuring that most users won't even notice something happened in between.

sinoru:
@rivera-ernesto Right. If create new target or anything else, It will make users notice it.

rivera-ernesto:
@sinoru Could you try to make a pull request for that?

sinoru:
@rivera-ernesto Okay. I make it right now.

rivera-ernesto:
And thanks for pointing out this issue @nrbrook!

sinoru:
I just create new pull request (#690)

rivera-ernesto:
@nrbrook Please give #660 a try and tell us if that works for you.

PR #674, opened 2015-12-24
fix bug: When log file rolling so fast (set maximumFileSize 100), fil…

glorybird:
…e create date may be same, may be leading to the newly created log file will be deleted

case:
fileLogger.maximumFileSize = 100;  // very small
fileLogger.rollingFrequency = 0;
fileLogger.logFileManager.maximumNumberOfLogFiles = 1;

you can add assert like following:
image

file log rolled very fast, lead to the same file creation time, sorting is not reliable.
may be leading to the newly created log file will be deleted

### DIFF
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -59,6 +59,7 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
     NSUInteger _maximumNumberOfLogFiles;
     unsigned long long _logFilesDiskQuota;
     NSString *_logsDirectory;
+    NSDate* _lastLogFileCreateDate;
 #if TARGET_OS_IPHONE
     NSString *_defaultFileProtectionLevel;
 #endif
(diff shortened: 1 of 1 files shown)


Classes/DDFileLogger.m
```diff
@@ -433,6 +434,28 @@ - (NSString *)newLogFileName {
...
+{
+    NSDate* fileCreateDate = [NSDate date];
+    if (_lastLogFileCreateDate == nil) {
+        DDLogFileInfo *logFileInfo = [self sortedLogFileInfos].firstObject;
+        if (logFileInfo) {
+            _lastLogFileCreateDate = logFileInfo.creationDate;
+        }
+    }
+    if (_lastLogFileCreateDate && (int)_lastLogFileCreateDate.timeIntervalSince1970 >= (int)fileCreateDate.timeIntervalSince1970) {
+        fileCreateDate = [_lastLogFileCreateDate dateByAddingTimeInterval:1];
```
rivera-ernesto: If you're creating so many log files `fileCreateDate + 1` may also collision so this won't be very useful.

Also I wonder why do you need so many log files. The number will be huge if you have files created at the exact same `NSDate`. Maybe you should use a larger granularity or log less?

glorybird: Thanks rivera-ernesto prompt reply，but this situation's real place in our projects, my example in order to confirm and amplify the emergence of possible.
when log files rapid rolled, 1 second generates two files are very possible, file creation time is very likely the same.
Please solve this problem, makes the code more robust.
You can not assume that the user how to use your library.

rivera-ernesto: Yes, I understand that there may be some edge cases as this, but your pull request does not fix the problem. As I said:

> If you're creating so many log files fileCreateDate + 1 may also collision so this won't be very useful.

glorybird: Each time a file is created, will be compared with _lastLogFileCreateDate, fileCreateDate always larger than _lastLogFileCreateDate. Finally fileCreateDate assigned to _lastLogFileCreateDate. Why also collision?

bpoplauschi:
@rivera-ernesto what's the situation with this PR? Does it require more work? Is this something we don't want?

rivera-ernesto:
My opinion was https://github.com/CocoaLumberjack/CocoaLumberjack/pull/674#r48479508

Classes/DDFileLogger.m
bpoplauschi: @glorybird any response to the previous statement?

rivera-ernesto:
Closing this for now. Feel free to continue the discussion and reopen if needed.

PR #693, opened 2016-02-24
Typo in import statement

maerki:
Replace LUmberjack with Lumberjack. Otherwise DDFileLogger.h is not found on case-sensitive file systems.


bpoplauschi:
Thanks @maerki

PR #702, opened 2016-03-22
Swift2.2

MarkQSchultz:
__FUNCTION__, __LINE__, __FILE__ have been replaced with #function, #line, and #file in Swift 2.2


rivera-ernesto:
As this is part of Xcode 7.3 I guess it breaks backwards compatibility right?
If so we better make a separate branch for it.

kyro38:
Yeah, some branching will probably be required to maintain the backward compatibility.

rivera-ernesto:
Ok, created a `Xcode_7.3` branch. Could you pull request there please?

MarkQSchultz:
Will do.

bpoplauschi:
It's great that you guys did the work on the `Xcode_7.3` branch and I think that in order for us to be ok with semver, we need to release this change in a major release (i.e. 3.x) as it does break backwards compatibility.

Unfortunately, we can't rely on [#if swift](https://github.com/apple/swift-evolution/blob/master/proposals/0020-if-swift-version.md) as it only works in Swift 2.2 and later, but we will in the future.
@rivera-ernesto thoughts?

rivera-ernesto:
I don't think we need a major release for this as there's nothing complicated to do for users besides upgrading to Xcode 7.3.

bpoplauschi:
Yes, but it won't compile for existing users with their current setup.

rivera-ernesto:
Yes, but for that a separate branch is enough for now. Soon Xcode 7.3 will be mandatory for new submissions and the new branch will work for anyone.

antoinelamy:
Still no official version for this?

rivera-ernesto:
I think Xcode 7.3 is not mandatory yet.

PR #704, opened 2016-03-23
Update to Swift 2.2: Change `__FUNCTION__` to `#function`, `__FILE__` to `#file`

MarkQSchultz:
and `__LINE__` to `#line`

### DIFF
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -55,7 +55,7 @@ public func resetDefaultDebugLevel() {
 }
 
 @available(*, deprecated, message="Use one of the DDLog*() functions if appropriate or call _DDLogMessage()")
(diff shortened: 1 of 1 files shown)


MarkQSchultz:
Ref #702

rivera-ernesto:
Thanks!

rivera-ernesto:
I think we could release this even before Xcode 7.3 becomes mandatory. People with older version could just keep using the older version.

What do you think @bpoplauschi ?

MaddTheSane:
Agreed. Also note that the old-style `__FILE__`, etc… won't work on Swift 3.

bpoplauschi:
Not sure what to do here. We should see what other projects have done.

bpoplauschi:
I see Alamofire has switched to Xcode 7.3 as a requirement since version 3.3.0. I think we can do the same. Release a new version that requires Swift 2.2+ and Xcode 7.3+.

bpoplauschi:
Guys, I have created a 2.3.0 release with all the latest changes, that requires Xcode 7.3 and Swift 2.2. Hope this helps

rivera-ernesto:
Thanks a lot!

PR #723, opened 2016-04-28
replaced macros to support swift 3 styling. Removes warnings

tbass134:
Swift 3 replaces support for C macros for newer swift macros. This PR replaces these with the new swift 3 style, therefore, fixing the warnings in Xcode
see https://github.com/apple/swift-evolution/blob/master/proposals/0028-modernizing-debug-identifiers.md 
for more info on removal of C style macros


rivera-ernesto:
#704 duplicate.

PR #736, opened 2016-05-19
Ability to create new log files every day

anivaros:
I've added flag, that enables every day file creating.
This useful if you want to search logs for concrete day.

Without that flag messages for today may be logged to yesterday file.
With flag in one file will be messages only for concrete day.


Classes/DDFileLogger.m
```diff
@@ -47,7 +47,8 @@
 #endif
 
 unsigned long long const kDDDefaultLogMaxFileSize      = 1024 * 1024;      // 1 MB
-NSTimeInterval     const kDDDefaultLogRollingFrequency = 60 * 60 * 24;     // 24 Hours
+NSTimeInterval     const kDDOneDayTimeInterval         = 60 * 60 * 24;	   // 24 Hours
+NSTimeInterval     const kDDDefaultLogRollingFrequency = kDDOneDayTimeInterval;
```
rivera-ernesto: There may be some constants for this in the iOS SDK.

Classes/DDFileLogger.m
```diff
@@ -869,6 +874,32 @@ - (void)maybeRollLogFileDueToSize {
...
+	if (_rollEveryDay && _currentLogFileInfo && ![self isLogFileCreatedToday:_currentLogFileInfo]) {
+		NSLogVerbose(@"DDFileLogger: Rolling log file due to date...");
+		
+		[self rollLogFileNow];
+	}
+}
+
+- (BOOL)isLogFileCreatedToday:(DDLogFileInfo *)aLogFile {
+	NSDate *logFileCreationDate = aLogFile.creationDate;
+	NSDate *today = [NSDate date];
```
rivera-ernesto: This depends on the time zone of the user. So doesn't have much more meaning than rolling every 24 hours as in the current implementation.

anivaros: Yes, it depends. Method comparing dates according to **current** calendar, see screenshot. 
screen shot 2016-05-23 at 21 45 31

rivera-ernesto: > With my flag this will not happen. in file "2016-05-23.log" will be messages only for 23, not for 24.

Depending on the time zone that is.

I think that if you need something very specific you could subclass the file logger and override the roll methods exactly to what you need. You will have all the flexibility to modify the behavior.

anivaros: I think, problems will be only if user moving to different timezones due the day, but in that case time in logged messages will be strange too.

File logger class has no template methods, or another ways to override rolling without pain.
If I could, I've did that already. But to achieve same behavior I should rewrite private methods, htat not cool, because they could be changed someday. And my logger become broken.

rivera-ernesto: I prefer that you modify the file logger so people can easily override the rolling methods. You can make it public and easier to override in a PR.

anivaros: > I prefer that you modify the file logger so people can easily override the rolling methods. 

In last commit I did that:
Added "willlog" and "didLog" methods and added "shouldArchive" method

anivaros: Additionally I've made _currentLogFileInfo instance var protected to use it in subclasses.

anivaros:
Hello again!

> I prefer that you modify the file logger so people can easily override the rolling methods.
> In last commit I did that:

Added "willlog" and "didLog" methods and added "shouldArchive" method

Additionally I've made _currentLogFileInfo instance var protected to use it in subclasses.

PR #747, opened 2016-06-22
Pin CocoaLumberjackSwift to Swift 2.3 to allow compilation on Xcode 8.

jbmorley:
CocoaLumberjackSwift is not currently compatible with Swift 3. This pins the version of Swift to version 2.3 to allow the current code to be safely compiled using both Xcode 7 and Xcode 8 beta.


jbmorley:
@rivera-ernesto Thanks! 👍

PR #750, opened 2016-06-28
To adapted swift

harryzjm:
I had add some func just like swift print func
so , we can easy to use like this:
`DLogInfo("1", 1, [2,3,4])`


danielgalasko:
Don't see the value in this as a PR, it sounds far too specific to your personal use case

bpoplauschi:
I agree with @danielgalasko this is not something should be included in the library. You can keep using it from your own fork.

PR #751, opened 2016-07-01
Support Swift 3.0

iPersona:
Just fix issues when compiling on Xcode8 Swift 3.0


johndpope:
It might be necessary to recode  like this -> 
https://github.com/nvzqz/Sage/blob/master/Sources/Game.swift

   ```
 /// A chess game mode.
    public enum Mode {

        #if swift(>=3)

        /// A game between two humans.
        case humanVsHuman

        /// A game between a human and a computer.
        case humanVsComputer

        /// A game between two computers.
        case computerVsComputer

        #else

        /// A game between two humans.
        case HumanVsHuman

        /// A game between a human and a computer.
        case HumanVsComputer

        /// A game between two computers.
        case ComputerVsComputer

        #endif

    }
```

PR #754, opened 2016-07-25
Use proper supported platforms identifiers

dulaccc:
Small fix to use the proper identifiers on the Framework targets.
It helps third party tools correctly identify the platforms too.

Cheers


rivera-ernesto:
Can you please fix the conflicts?

bpoplauschi:
@dulaccc thanks, but I replaced your PR with 40f1852 - we don't even need to explicitly set `SUPPORTED_PLATFORMS` - they are picked up from the system default based on the SDK

Issue #766, opened 2016-08-30
Won't compile with Carthage due to SWIFT_VERSION in Xcode 8 beta 6


lvandal:
I've reported this a week ago and then the issue was closed even though this is still a problem. The project needs to be updated in order to compile with the latest Xcode 8 beta.

###Steps to reproduce

- Try to compile CocoaLumberJack via Carthage

###Output

=== BUILD TARGET CocoaLumberjackSwift OF PROJECT Lumberjack WITH CONFIGURATION Release ===

Check dependencies
“Use Legacy Swift Language Version” (SWIFT_VERSION) is required to be configured correctly for targets which use Swift. Use the [Edit > Convert > To Current Swift Syntax…] menu to choose a Swift version or use the Build Settings editor to configure the build setting directly.
“Use Legacy Swift Language Version” (SWIFT_VERSION) is required to be configured correctly for targets which use Swift. Use the [Edit > Convert > To Current Swift Syntax…] menu to choose a Swift version or use the Build Settings editor to configure the build setting directly.

danielgalasko:
Can you link the original issue you reported? So weird that there is no word on Swift3 support, not even an experimental branch ¯\_(ツ)_/¯

bpoplauschi:
Should be fixed now by #747 and #773

PR #767, opened 2016-09-01
Updated project to use latest Swift 3

asynchrony-ringo:

* [X] I have added the required tests to prove the fix/feature I am adding
* [NA] I have updated the documentation (if necesary)
* [ ] I have run the lint and it passes (`pod lib lint`) **In fact, using the new beta of cocoapods, which is required to work in Swift 3, a non-changed value does seem to cause an error, but this may be erroneous. It also majorly screws up the local git checkout to run that command in the beta.** 

This merge request fixes / refers to the following issues: #766 

### Pull Request Description

Converted the project to support Swift 3 beta 6. Many open-source projects are choosing to do this in a "swift3" branch for the time being. This pull request is against master, but it may be preferable to create a swift3 branch and merge the pull request into that instead.

### DIFF
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -16,7 +16,7 @@
 import Foundation
 
 extension DDLogFlag {
-    public static func fromLogLevel(logLevel: DDLogLevel) -> DDLogFlag {
+    public static func fromLogLevel(_ logLevel: DDLogLevel) -> DDLogFlag {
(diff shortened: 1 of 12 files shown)


danielgalasko:
Would it not be a good idea to add Swift 2.3 support as well so we dont force all consumers to use Swift 3...

bpoplauschi:
Many thanks @asynchrony-ringo, but #769 was more comprehensive and I merged that one, it contains everything we need for Swift 3.0 and Xcode 8.

PR #745, opened 2016-06-09
Fix CLIColor.h not found for non-AppKit binaries w/o clang modules

thibault-ml:
### New Pull Request Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](http://cocoadocs.org/docsets/CocoaLumberjack/)
* [x] I have searched for a similar pull request in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/pulls) and found none

* [x] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)
* [x] I have run the tests and they pass

This merge request fixes / reffers to the following issues: #744 

### Pull Request Description

This pull requests adds CLIColor.h/m to the OS X target and makes sure the header file is shipped with the Framework so that apps without AppKit (or using ```DD_CLI```) that also disable Clang modules can actually use CocoaLumberjack.

PR #781, opened 2016-09-27
Expose CLIColor header publicly in cross-platform framework target

erichoracek:

### Pull Request Description

...

In #745, CLIColor.h was made public. However, there was not a corresponding addition to CocoaLumberjack.h to make this file publicly available. This should fix that issue.

This results in the following error when attempting to import CocoaLumberjack (when built via Carthage):
```
Umbrella header for module 'CocoaLumberjack' does not include header 'CLIColor.h'
```

### DIFF
--- a/Framework/Lumberjack/CocoaLumberjack.h
+++ b/Framework/Lumberjack/CocoaLumberjack.h
@@ -37,3 +37,6 @@ FOUNDATION_EXPORT const unsigned char CocoaLumberjackVersionString[];
 #import <CocoaLumberjack/DDTTYLogger.h>
 #import <CocoaLumberjack/DDASLLogger.h>
 #import <CocoaLumberjack/DDFileLogger.h>
+
+// CLI
+#import <CocoaLumberjack/CLIColor.h>


erichoracek:
cc @thibault-ml @rivera-ernesto

diederich:
I just stumbled over this as well. So technically the CLIColor.h should not be needed in all frameworks (e.g. not in the iOS and WatchOS from what I've seen), but it seems like this doesn't work well with Xcode. 
When I explicitly specify which target to build in my scheme (CocoaLumberjackSwift-iOS), I can see the built framework and it doesn't include CLIColor.h in the Headers folder. That is expected and then there is no error about it missing from the umbrella header.
The problem comes once I specify the 'generic'  CocoaLumberjack.framework. E.g. when I add it to the binary's embedded frameworks list, Xcode seems to pick up the 'wrong' framework target and runs the copy headers phase for the main target, which adds the CLIColors.h header to the Headers folder in the framework and on the next rebuild it would complain about it missing from the umbrella header.

I think this can be solved in 2 ways:
1) Rename the targets (and from what it looks like the resulting framework as well) to include the platform, which would fix the CLIColor.h header showing up
2) Just include in in the umbrella header and for all of the builds

diederich:
@erichoracek Now that CLIColor.h is in the umbrella header, we should probably have it in all configurations public as well. Currently it's only in the OSX framework:
screen shot 2016-10-03 at 14 12 31
(ignore the -OSX, I've just played with this to see if I can find another workaround)

erichoracek:
Unfortunately this won't work, as the "CocoaLumberjack" framework is not macOS only, rather it is a multi-architecture framework as described here http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/

erichoracek:
@diederich There are two umbrella headers—one for ios/watch/tvOS, while the other is for the [cross-platform framework](http://ilya.puchka.me/xcode-cross-platform-frameworks/). I've only added `CLIColor.h` to the cross-platform framework because that is the only one that includes `CLIColor`. 

It sounds like you may additionally experiencing a separate issue caused by the fact that Xcode is unable to choose which framework to build due to the fact that there are two targets that produce an framework for iOS (see https://github.com/CocoaLumberjack/CocoaLumberjack/issues/782). When you're using Xcode's "Find implicit dependencies" setting, it won't be able to disambiguate which you want, since `Lumberjack.xcodeproj` has produces two iOS frameworks named `CocoaLumberjack`.

ronak2121:
Will this be merged and version 3.0.1 be released soon? This is blocking us and will force us to patch on our end (which we'd like to avoid).

VictorVanDoom:
Is there a ETA for this fix yet? We too are currently blocked by this issue, as it prevents our build-server to properly run builds and tests of our Apps.

max-potapov:
@VictorVanDoom This fix is already merged to master. We need to ask @rivera-ernesto to bump version or you could try use latest version from master branch.

rivera-ernesto:
I don't feel confident about making releases as I don't actively use the library right now.

Either ask @bpoplauschi or you can always [force](https://guides.cocoapods.org/syntax/podfile.html#pod) CocoaPods to use the latest master.

VictorVanDoom:
@max-potapov Are you sure that the fix works correctly? I've done quite some testing now and the error still occurs when building for iOS-Swift.

PR #796, opened 2016-11-02
Conditionally include CLIColor in umbrella header

brendonjustin:

* n/a I have added the required tests to prove the fix/feature I am adding
* n/a I have updated the documentation (if necesarry)
* [ ] I have run the tests and they pass **:( see below**

This merge request fixes / reffers to the following issues: <none>

I get linker errors if I try to run any of the tests, even on master, so I can't test this :(. The compiled framework with this change does work in my app the same as one without the change, at least.

### Pull Request Description

`CocoaLumberjack.h`, what seems to be the include-everything-public header for CocoaLumberjack, includes `CLIColor.h` for all platforms. But `CLIColor.h` is not added as a public header in the project settings for the iOS/watchOS/tvOS targets, e.g. `CocoaLumberjack-iOS`, and therefore is not included in the umbrella header of the resulting frameworks. Including one of those frameworks in an iOS project results in Xcode complaining with a warning like this:
```
/[snip]/<module-includes>:1:1: Umbrella header for module 'CocoaLumberjack' does not include header 'CLIColor.h'
```

I wasn't sure exactly what the fix for this is. I took a stab at it but wouldn't be surprised if this isn't the right approach.

This issue applies to CocoaLumberjack 3.0.0 and at least back to 2.4.0.

### DIFF
--- a/Framework/Lumberjack/CocoaLumberjack.h
+++ b/Framework/Lumberjack/CocoaLumberjack.h
@@ -39,4 +39,6 @@ FOUNDATION_EXPORT const unsigned char CocoaLumberjackVersionString[];
 #import <CocoaLumberjack/DDFileLogger.h>
 
 // CLI
+#if defined(DD_CLI) || !__has_include(<AppKit/NSColor.h>)
 #import <CocoaLumberjack/CLIColor.h>
+#endif


thibault-ml:
This broke CocoaLumberjack for me :-/ I have to uncomment those changes for it to compile, otherwise I get the following error:

screen shot 2016-12-02 at 17 04 02

rivera-ernesto:
Reverted it for now.

Issue #815, opened 2016-12-09
Umbrella header for module 'CocoaLumberjack' does not include header 'CocoaLumberjack.common-umbrella.h'


arifken:
### New Issue Checklist

* [X] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [X] I have read the [Documentation](https://github.com/CocoaLumberjack/CocoaLumberjack#documentation) or [CocoaDocs](cocoadocs.org/docsets/CocoaLumberjack)
* [X] I have searched for a similar issue in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/issues) and found none

### Issue Info

Hello! I'm getting a compile error when trying to use CocoaLumberjack in a WatchKit Extension (WatchOS 1.0, unfortunately). For some reason, it was working in version 2.3.0 of CocoaLumberjack, but in version 3.0.x I get the following compile error:

[project path]/<module-includes>:1:1: Umbrella header for module 'CocoaLumberjack' does not include header 'CocoaLumberjack.common-umbrella.h'

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform Name           | ios / watchos
 Platform Version        | 10.1
 CocoaLumberjack Version | 3.0.0
 Integration Method      | cocoapods
 Xcode Version           | e.g. Xcode 8.1
 Repro rate              | 100% of the time

### Issue Description and Steps

[project path]/<module-includes>:1:1: Umbrella header for module 'CocoaLumberjack' does not include header 'CocoaLumberjack.common-umbrella.h'

This error only seems to come up when I do an Archive. If i just build/run the same configuration it works. Any ideas?

Thanks!

### UPDATE:

I was able to work around the issue by adding these two lines back into CocoaLumberjack.podspec:

```
s.public_header_files = 'Classes/*.h'

s.module_map = 'Framework/Lumberjack/CocoaLumberjack.modulemap'
```

sushichop:
@arifken 

Hi! Could you try to use master branch? 
```
target 'YourTarget' do 
  use_frameworks!
  pod 'CocoaLumberjack', :git => 'git@github.com:CocoaLumberjack/CocoaLumberjack.git'
end
```
As far as I understand, the podspec of CocoaLumberjack 3.0.0 has problem about umbrella header.
It doesn't work properly especially when you use Objective-C.

[Note]
[...]

PR #824, opened 2017-01-11
use custom modulemap

sushichop:

* [ ] I have updated the documentation (if necesarry)

This merge request reffers to the following issue: #815 
This merge request reffers to the following commit: https://github.com/CocoaLumberjack/CocoaLumberjack/commit/0f5a793843c748c06511e5c3d1e20967798b50d1

### Pull Request Description

Custom module map was removed by this commit(https://github.com/CocoaLumberjack/CocoaLumberjack/commit/0f5a793843c748c06511e5c3d1e20967798b50d1) and it should be a temporizing workaround. This PR enables custom modulemap again as CocoaLumberjack used it before(e.g. CocoaLumberjack 2.3.0 used it.).

However, keep in mind. 
In order to do this, podspec will be summarized to 3 subspecs for the reason why latest CocoaPods (probably from version 1.0.0) requires the strict podspec. 

For example,` DDDispatchQueueLogFormatter` is declared explicitly in the CocoaLumberjack.modulemap file, on the other hand, the original several subspecs(`Core` or `Extensions`) does not include this header. In this case `pod lib lint` fails. In other words, it is a root cause.
Moreover, as a podspec syntax, you can not specify `module_map` attribute in subspec level.
So I needed to change podspec.

Thanks.

### DIFF
--- a/CocoaLumberjack.podspec
+++ b/CocoaLumberjack.podspec
@@ -24,22 +24,16 @@ Pod::Spec.new do |s|
   s.watchos.deployment_target = '2.0'
   s.tvos.deployment_target    = '9.0'
 
-  s.default_subspecs = 'Default', 'Extensions'
+  s.module_map = 'Framework/Lumberjack/CocoaLumberjack.modulemap'
+  s.default_subspecs = 'Default'
 
   s.subspec 'Default' do |ss|
-    ss.source_files         = 'Classes/CocoaLumberjack.h', 'Classes/DD*.{h,m}'
-    ss.public_header_files  = 'Classes/CocoaLumberjack.h', 'Classes/DD*.h'
-  end
-
-  s.subspec 'Core' do |ss|
-    ss.source_files         = 'Classes/DD*.{h,m}'
-    ss.public_header_files  = 'Classes/DD*.h'
-  end
-
-  s.subspec 'Extensions' do |ss|
-    ss.dependency 'CocoaLumberjack/Default'
-    ss.source_files         = 'Classes/Extensions/*.{h,m}'
-    ss.public_header_files  = 'Classes/Extensions/*.h'
+    ss.source_files         = 'Classes/CocoaLumberjack.h',
+                              'Classes/DD*.{h,m}',
+                              'Classes/Extensions/*.{h,m}'
+    ss.public_header_files  = 'Classes/CocoaLumberjack.h',
+                              'Classes/DD*.h',
+                              'Classes/Extensions/*.h'
   end
   
   s.subspec 'CLI' do |ss|


CocoaLumberjack.podspec
```diff
@@ -24,22 +24,16 @@ Pod::Spec.new do |s|
   s.watchos.deployment_target = '2.0'
   s.tvos.deployment_target    = '9.0'
 
-  s.default_subspecs = 'Default', 'Extensions'
+  s.module_map = 'Framework/Lumberjack/CocoaLumberjack.modulemap'
+  s.default_subspecs = 'Default'
```
rivera-ernesto: This may not be needed anymore but I'm not sure about losing the other subspecs.

sushichop: Thank you for your comment.

**s.module_map**

If you do not have a plan to support backward compatibility, 
`s.module_map` is not needed anymore.
In this case, backward compatibility means the followings.

When user uses `DDDispatchQueueLogFormatter` class
 [CocoaLumberjack 2.3]
import CocoaLumberjack.DDDispatchQueueLogFormatter
[CocoaLumberjack latest]
import CocoaLumberjack

To be honest, I also think cusutom modulemap is not needed anymore as you say😁
Without custom modulemap, CocoaPods automatically makes a normal modulemap file.
I made a PR to discuss about it.

**s.default_subspecs**
On the other hand, `s.default_subspecs` should be needed.
Because if this setting is removed from podspec, Swift Standard Libraries are always included even when user wants to use only Objective-C code.

To sum it up, I'll be gladly to decline(close) this PR if you do not have a plan to support backward compatibility about custom modulemap.

BTW, if you do not need custom modulemap, I think probably you had better change the modulemap of Lumberjack.xcodeproj a little. Each target still uses custom modulemap file that is not different from a modulemap file CocoaPods automatically makes.
Can I make an another PR to do this?

rivera-ernesto: Yes please. I think that at this point backwards compatibility with 2.x.x is less important than having cleaner project.

sushichop: Good! Then I'm closing this PR in first. About another PR..., please wait for a little while? (probably at the longest less than 1 week though)
Thanks!

sushichop: @rivera-ernesto 

> I think probably you had better change the modulemap of Lumberjack.xcodeproj a little.

I checked deeply and **actually** each target's headers and modulemap. As a result  I found no need change... Sorry for confusing you...
But if I find another issue or come to better solution, I'd like to discuss or make a PR.
Thanks again!!

PR #829, opened 2017-02-07
Optimized USE_DISPATCH_CURRENT_QUEUE_LABEL and USE_DISPATCH_GET_CURRENT_QUEUE macros.

jcbertin:

### Pull Request Description

I propose to optimize OS detection in `USE_DISPATCH_CURRENT_QUEUE_LABEL` and `USE_DISPATCH_GET_CURRENT_QUEUE`macros by using `dispatch_once_f()` call on iOS and macOS. I use `dispatch_once_f()` instead of `dispatch_once()` to save more CPU cycles.
I wonder if it could optimize `USE_PTHREAD_THREADID_NP` as well: this would replace one double comparison by two boolean comparisons in the best case scenario.

### DIFF
--- a/Classes/DDLog.m
+++ b/Classes/DDLog.m
@@ -21,6 +21,7 @@
 #import "DDLog.h"
 
 #import <pthread.h>
+#import <dispatch/dispatch.h>
 #import <objc/runtime.h>
 #import <mach/mach_host.h>
 #import <mach/host_info.h>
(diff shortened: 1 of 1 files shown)


Classes/DDLog.m
```diff
@@ -1002,15 +1003,30 @@ @implementation DDLogMessage
...
+  #define USE_DISPATCH_GET_CURRENT_QUEUE   (_dispatch_queue_label_init(), _use_dispatch_get_current_queue)
 
 #elif TARGET_OS_WATCH || TARGET_OS_TV
 
 // Compiling for watchOS, tvOS
 
-#define USE_DISPATCH_CURRENT_QUEUE_LABEL YES
-#define USE_DISPATCH_GET_CURRENT_QUEUE   YES
+  #define USE_DISPATCH_CURRENT_QUEUE_LABEL YES
+  #define USE_DISPATCH_GET_CURRENT_QUEUE   NO
```
jcbertin: Let the compiler optimize this test branch.

rivera-ernesto: Not a big deal, but why do we need this white space?

jcbertin: For alignment with other macros…

PR #831, opened 2017-02-08
Crash  [DDLog log:level:flag:context:file:function:line:tag:format:]

jack3010:
Fix crash for ios 9 and 10 

### Pull Request Description

We need the nil or some other way to know how many arguments to grab. `va_list` has no magic way to know when to stop.

Things like `stringWithFormat:` don't need a nil because it determines the number of arguments based on the number of format specifiers (which is why they need to match or your code goes boom). 
What you could do is make use of the following NSString method:
```
va_start(args, format);
    NSString *msg = [[NSString alloc] initWithFormat:format arguments:args];
    // do whatever
va_end(args);
```
Please help me to review.

### DIFF
--- a/Classes/DDLog.m
+++ b/Classes/DDLog.m
@@ -381,6 +381,11 @@ static NSUInteger _numProcessors;
         va_start(args, format);
         
         NSString *message = [[NSString alloc] initWithFormat:format arguments:args];
+        
+        va_end(args);
+        
+        va_start(args, format);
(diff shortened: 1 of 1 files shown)


kingandyoga:
Hi, I meet the same crash like yours in my own code. Your solution help me a lot, but I still don't understand what leads to the crash. Could you explain more clearly? Helps a lot~

PR #832, opened 2017-02-11
Added option to not copy messages.

jcbertin:

### DIFF
--- a/Classes/DDLog.m
+++ b/Classes/DDLog.m
@@ -1107,15 +1107,16 @@ static __inline__ __attribute__((__always_inline__)) void _dispatch_queue_label_
                         options:(DDLogMessageOptions)options
                       timestamp:(NSDate *)timestamp {
     if ((self = [super init])) {
-        _message      = [message copy];
+        BOOL copyMessage = (options & DDLogMessageDontCopyMessage) == 0;
+        _message      = copyMessage ? [message copy] : message;
         _level        = level;
         _flag         = flag;
         _context      = context;
 
-        BOOL copyFile = (options & DDLogMessageCopyFile) == DDLogMessageCopyFile;
+        BOOL copyFile = (options & DDLogMessageCopyFile) != 0;
         _file = copyFile ? [file copy] : file;
 
-        BOOL copyFunction = (options & DDLogMessageCopyFunction) == DDLogMessageCopyFunction;
+        BOOL copyFunction = (options & DDLogMessageCopyFunction) != 0;
         _function = copyFunction ? [function copy] : function;
 
         _line         = line;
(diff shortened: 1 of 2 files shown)


Classes/DDLog.m
```diff
@@ -1107,15 +1107,16 @@ - (instancetype)initWithMessage:(NSString *)message
...
     if ((self = [super init])) {
-        _message      = [message copy];
+        BOOL copyMessage = (options & DDLogMessageDontCopyMessage) == 0;
+        _message      = copyMessage ? [message copy] : message;
         _level        = level;
         _flag         = flag;
         _context      = context;
 
-        BOOL copyFile = (options & DDLogMessageCopyFile) == DDLogMessageCopyFile;
+        BOOL copyFile = (options & DDLogMessageCopyFile) != 0;
```
jcbertin: It's always quicker to compare against zero.

Classes/DDLog.m
```diff
@@ -1107,15 +1107,16 @@ - (instancetype)initWithMessage:(NSString *)message
                         options:(DDLogMessageOptions)options
                       timestamp:(NSDate *)timestamp {
     if ((self = [super init])) {
-        _message      = [message copy];
+        BOOL copyMessage = (options & DDLogMessageDontCopyMessage) == 0;
+        _message      = copyMessage ? [message copy] : message;
```
rivera-ernesto: If we want performance then we probably don't need an extra variable and can just do:

`_message      = (options & DDLogMessageDontCopyMessage) ? message : [message copy];`

jcbertin: IMHO, I first thought that the compiler was smart enough to produce to the same code. But since I wanted to be sure, I double-checked by using a disassembler, and guess what? The code is the same! You can verify it by looking at the attachments, line 49-51: the first file is with the explicit variable, the second is your version.
For my part, I prefer the first version since it looks like the code already there.

[iOSdm Daemon_-[DDLogMessage initWithMessage:level:flag:context:file:function:line:tag:options:timestamp:].txt](https://github.com/CocoaLumberjack/CocoaLumberjack/files/771654/iOSdm.Daemon_-.DDLogMessage.initWithMessage.level.flag.context.file.function.line.tag.options.timestamp.txt)
[iOSdm Daemon_-[DDLogMessage initWithMessage:level:flag:context:file:function:line:tag:options:timestamp:]_2.txt](https://github.com/CocoaLumberjack/CocoaLumberjack/files/771655/iOSdm.Daemon_-.DDLogMessage.initWithMessage.level.flag.context.file.function.line.tag.options.timestamp._2.txt)

jcbertin: BTW, the code is the same only with optimizations activated…

rivera-ernesto: Interesting.
Thanks!

PR #836, opened 2017-02-15
Added new hooks when adding loggers and formatters.

jcbertin:

### Pull Request Description

Problem: if a logger doesn't inherit from DDAbstractLogger and doesn't provide a queue (i.e..-loggerQueue returns nil), DDLog will create a queue for it. However it doesn't know anything about this queue and cannot create an environment in it. For formatters, the problem is the same: they don't know on which logger queue they are being added.
I propose this new API to compensate this problem. Maybe we should mark the old notifiers as obsolete…

### DIFF
--- a/Classes/DDLog.h
+++ b/Classes/DDLog.h
@@ -593,6 +593,19 @@ NSString * DDExtractFileNameWithoutExtension(const char *filePath, BOOL copy);
 - (void)didAddLogger;
 
 /**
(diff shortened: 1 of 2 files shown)


jcbertin:
What do you think of marking -[DDLogger didAddLogger] and -[DDLogFormatter didAddToLogger:] as deprecated?

rivera-ernesto:
I think that's safe if we still keep the deprecated methods around for some time.

jcbertin:
ok!

PR #867, opened 2017-04-28
Fixed debug symbols stripping

radulov:

### Pull Request Description

This pull request changes allow to strip debug symbols for framework users.
Stripping debug symbols decrease application size.
...


rivera-ernesto:
AFAIK debug symbols will get stripped from AppStore builds automatically.

On the other hand removing the symbols won't allow debugging for developers.

radulov:
Developers use debug configuration, so they have ability to check DSYMs. For custom release builds need to strip all debug symbols in case of its useless.

rivera-ernesto:
To analyze Release/Production crash reports you still need to upload those symbols to your crash report service.

If you check your release binaries, symbols (dSYM) files are never included in there. So they don't account into your app's size.

radulov:
Yes, dSYM files never included in release build, but framework release binaries contains debug symbols. You can easy check it using command "nm". For example output(release binary) before my changes:
[Before.txt](https://github.com/CocoaLumberjack/CocoaLumberjack/files/965222/Before.txt)
and after:
[After.txt](https://github.com/CocoaLumberjack/CocoaLumberjack/files/965221/After.txt)

rivera-ernesto:
> but framework release binaries contains debug symbols

Which isn't bad at all. You get the smaller final app and have the ability to analyze framework's crashes.

radulov:
But how binaries with additional symbols can be smaller? And even without debug symbols we can analyze crashes in release builds, for example in Fabric system. At the same time, crash log for end users of our products will not contain real names.

rivera-ernesto:
I think we discussed everything already:

- You're app won't be bigger because of symbols.
- You need symbols to analyze production crashes.
- End users won't have the symbols, so crashes won't contain "real names".

Closing this for now.

PR #871, opened 2017-05-21
Update for Swift 3.1

farice:
Mainly syntactical changes.

### DIFF
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -16,7 +16,7 @@
 import Foundation
 
 extension DDLogFlag {
-    public static func fromLogLevel(logLevel: DDLogLevel) -> DDLogFlag {
+    public static func fromLogLevel(_ logLevel: DDLogLevel) -> DDLogFlag {
         return DDLogFlag(rawValue: logLevel.rawValue)
     }
     
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -16,28 +16,30 @@
 import Foundation
 
 extension DDLogFlag {
-    public static func from(_ logLevel: DDLogLevel) -> DDLogFlag {
+    public static func fromLogLevel(logLevel: DDLogLevel) -> DDLogFlag {
         return DDLogFlag(rawValue: logLevel.rawValue)
     }
-	
-	public init(_ logLevel: DDLogLevel) {
+    
+    public init(_ logLevel: DDLogLevel) {
         self = DDLogFlag(rawValue: logLevel.rawValue)
-	}
+    }
     
     ///returns the log level, or the lowest equivalant.
     public func toLogLevel() -> DDLogLevel {
-        if let ourValid = DDLogLevel(rawValue: rawValue) {
+        if let ourValid = DDLogLevel(rawValue: self.rawValue) {
             return ourValid
         } else {
-            if contains(.verbose) {
+            let logFlag:DDLogFlag = self
+            
+            if logFlag.contains(.verbose) {
                 return .verbose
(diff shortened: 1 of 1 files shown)


Classes/CocoaLumberjack.swift
```diff
@@ -52,39 +54,44 @@ public func resetDefaultDebugLevel() {
...
-    _DDLogMessage(message, level: level, flag: .verbose, context: context, file: file, function: function, line: line, tag: tag, asynchronous: async, ddlog: ddlog)
+public func DDLogVerbose( message: @autoclosure () -> String, level: DDLogLevel = defaultDebugLevel, context: Int = 0, file: StaticString = #file, function: StaticString = #function, line: UInt = #line, tag: AnyObject? = nil, asynchronous async: Bool = true, ddlog: DDLog = DDLog.sharedInstance()) {
+    _DDLogMessage(message: message, level: level, flag: .verbose, context: context, file: file, function: function, line: line, tag: tag, asynchronous: async, ddlog: ddlog)
 }
 
-public func DDLogError(_ message: @autoclosure () -> String, level: DDLogLevel = defaultDebugLevel, context: Int = 0, file: StaticString = #file, function: StaticString = #function, line: UInt = #line, tag: Any? = nil, asynchronous async: Bool = false, ddlog: DDLog = DDLog.sharedInstance) {
```
rivera-ernesto: Why remove unnamed parameters?

sushichop:
IMO, CocoaLumberjack 3.2.0 already supports Swift 3.1.

And why do we need to remove unnamed parameters as @rivera-ernesto  says ?
I think the unnamed parameter at 1st argument is for a common swifty style, so I think we should not remove it...

farice:
You are right. Unnamed parameters as a 1st argument is mainstream Swifty style. I reverted those changes.

Classes/CocoaLumberjack.swift
```diff
@@ -16,28 +16,30 @@
 import Foundation
 
 extension DDLogFlag {
-    public static func from(_ logLevel: DDLogLevel) -> DDLogFlag {
+    public static func fromLogLevel(_ logLevel: DDLogLevel) -> DDLogFlag {
```
rivera-ernesto: Agree that here `from(logLevel: DDLogLevel)` would be more "Swift".

Classes/CocoaLumberjack.swift
```diff
@@ -52,39 +54,44 @@ public func resetDefaultDebugLevel() {
...
+        ddlog.log(asynchronous, message: logMessage)
     }
 }
 
-public func DDLogDebug(_ message: @autoclosure () -> String, level: DDLogLevel = defaultDebugLevel, context: Int = 0, file: StaticString = #file, function: StaticString = #function, line: UInt = #line, tag: Any? = nil, asynchronous async: Bool = true, ddlog: DDLog = DDLog.sharedInstance) {
-    _DDLogMessage(message, level: level, flag: .debug, context: context, file: file, function: function, line: line, tag: tag, asynchronous: async, ddlog: ddlog)
+public func DDLogDebug(_ message: @autoclosure () -> String, level: DDLogLevel = defaultDebugLevel, context: Int = 0, file: StaticString = #file, function: StaticString = #function, line: UInt = #line, tag: AnyObject? = nil, asynchronous async: Bool = true, ddlog: DDLog = DDLog.sharedInstance()) {
+    _DDLogMessage(message: message, level: level, flag: .debug, context: context, file: file, function: function, line: line, tag: tag, asynchronous: async, ddlog: ddlog)
```
rivera-ernesto: I think now `Any` is preferred over `AnyObject`.

rivera-ernesto:
3.3.0 now supports Swift 4.

PR #895, opened 2017-10-08
Fixes CI test failure(Xcode 9.0, Swift 4.0).

sushichop:

This merge request fixes CI test failure.

### Pull Request Description

I mainly fixed `pod lib lint` error/warning and `travis CI test` failure.
Now all tests should pass with Xcode 9.0 and Swift 4.0.


sushichop:
OK. CI test turns green😁

Classes/DDFileLogger.m
```diff
@@ -147,8 +147,8 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
                       ofObject:(id)object
                         change:(NSDictionary *)change
                        context:(void *)context {
-    NSNumber *old = change[NSKeyValueChangeOldKey];
-    NSNumber *new = change[NSKeyValueChangeNewKey];
+    NSNumber *old = [change objectForKey:NSKeyValueChangeOldKey];
+    NSNumber *new = [change objectForKey:NSKeyValueChangeNewKey];
```
bpoplauschi: why those changes and the similar ones?
You are basically going back from modern ObjC like `dict[key]` to `[dict objectForKey:key]`

sushichop: @bpoplauschi 

Good question because I wanted to discuss this point!
This change fixes warning about CLANG_WARN_UNGUARDED_AVAILABILITY that has been introduced from Xcode 9. Executing `pod lib lint` with Xcode 9, you will see many warnings.
By this new warning feature, if CocoaLumberjack wants to keep using modern ObjC like `dict[key]`, it must set minimum deployment target version macOS 10.8, iOS 6.0. 
On the other hand, if CocoaLumberjack wants to keep supporting macOS 10.7 and iOS 5.0 like podspec, it must go back to use traditional(old) ObjC style.

Related(modern ObjC) API is here. API_AVAILABLE macro seems to be newly added from Xcode 9.
`- (nullable ObjectType)objectForKeyedSubscript:(KeyType)key API_AVAILABLE(macos(10.8), ios(6.0), watchos(2.0), tvos(9.0));`

In other words, there are 3 options to fix CLANG_WARN_UNGUARDED warning. 
(1) CocoaLumberjack goes back to use traditional(old) ObjC style. This keeps [supporting macOS 10.7 and iOS 5.0](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/CocoaLumberjack.podspec#L22-L23) like podspec as before.
(2) CocoaLumberjack terminates [supporting macOS 10.7 and iOS 5.0](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/CocoaLumberjack.podspec#L22-L23), then it changes podspec to macOS 10.8 and iOS 6.0. This keeps using modern ObjC style in CocoaLumberjack.
(3) Just ignore CLANG_WARN_UNGUARDED warning.

This time I chose (1). However, at the same time, I think that it may be better to choose (2)...🤔
Thanks in advance.

bpoplauschi: @sushichop this makes perfect sense, thanks for the explanation.

I think we are safe to remove support for iOS 5 and macOS 10.7 in the new version, since we have enough versions that support them and they are with very low usage - most app devs will not target more than the max os version + 1 or 2 version before that

sushichop: @bpoplauschi 
Thanks for your comment. I reverted above commit(1ca6d0f) and set minimum deployment targets to macOS 10.8 and iOS 6.0.

> most app devs will not target more than the max os version + 1 or 2 version before that

**I fully agree with you**, but I set macOS 10.8 and iOS 6.0 **for now**.
If it is needed to update to newer versions, I'll send another pull request☺️
Is it all right?

sushichop:
OK. CI test passed again.

rivera-ernesto:
Thanks!

bpoplauschi:
👍

bpoplauschi:
This must be at least a minor release (3.4.0). I've updated the readme requirements and backwards compatibility with the iOS 5 and OS X 10.7 removal.

sushichop:
@bpoplauschi @rivera-ernesto 
Thanks for your checking.

PR #909, opened 2018-01-07
Fix DDLogFileManagerDefault -isLogFile:

ffried:

### Pull Request Description
Given the following two bundle identifiers:
- com.mycompany.app.abcservice
- com.mycompany.app.abcservice2

The latter would reuse/overwrite the log files of the former because `hasPrefix:` would return `YES`.
This fixes the problem by also checking for a space after the app name / bundle identifier.

### DIFF
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -279,7 +279,8 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
 - (BOOL)isLogFile:(NSString *)fileName {
     NSString *appName = [self applicationName];
 
-    BOOL hasProperPrefix = [fileName hasPrefix:appName];
+    // We need to add a space to the name as otherwise we could match applications that have the name prefix.
+    BOOL hasProperPrefix = [fileName hasPrefix:[appName stringByAppendingString:@" "]];
     BOOL hasProperSuffix = [fileName hasSuffix:@".log"];
     
     return (hasProperPrefix && hasProperSuffix);


ffried:
There's a little more to this fix than I previously thought. I'm adding this here just in case anyone else is wondering.

Given the following setup:
- An app with one or more extensions (Today Extension, Notification Extension, ...)
- They share a common app group for data sharing
- They all use CocoaLumberjack for logging
- They all use a FileLogger for logging into a file that lies within the common app group

Usually the extensions and the app share a common prefix in the bundle identifier:
- Main App: _com.mycompany.app_
- Extension 1: _com.mycompany.app.extension1_
- Extension 2: _com.mycompany.app.extension2_

Now, before this fix, you could ask the LogFileManager of the file logger in the app to return all log files (e.g. using `-unsortedLogFilePaths`) and you would get the log files of the app **and** those of the extensions that have the same prefix in the bundle identifier as the main app, too (Extension 1 and 2 in above's example).
But since `-unsortedLogFilePaths` etc. is using `-isLogFile:` to determine what files to include, this was actually incorrect behavior. To get **all** log files you should directly query the contents of the folder you passed to the log file manager when creating the file logger.

When I created this PR, I had the problem that some log files were re-used / overwritten. It took me some time just now to figure out this "side-effect". 
So hopefully I have now saved someone else's time. 🙂

ffried:
@bpoplauschi Thanks for the offer - which I gladly accept. Happy to help!

bpoplauschi:
Great @ffried. I will send you an invite.

Issue #918, opened 2018-03-16
DDAssert for Swift


ole:
### New Issue Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](https://github.com/CocoaLumberjack/CocoaLumberjack#documentation) or [CocoaDocs](cocoadocs.org/docsets/CocoaLumberjack)
* [x] I have searched for a similar issue in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/issues) and found none

### Issue Description and Steps

It seems there is currently no equivalent for `DDAssert` in CocoaLumberjack's Swift API. Would you be open to adding this? I thought I'd ask before I send a pull request.

I'd like to see two Swift functions, mimicking what's available in the Swift stdlib: [`DDAssert`](https://developer.apple.com/documentation/swift/1541112-assert) and [`DDAssertionFailure`](https://developer.apple.com/documentation/swift/1539616-assertionfailure).

I imagine the implementation would look something like this:

```swift
public func DDAssert(_ condition: @autoclosure () -> Bool, _ message: @autoclosure () -> String = "", level: DDLogLevel = defaultDebugLevel, context: Int = 0, file: StaticString = #file, function: StaticString = #function, line: UInt = #line, tag: Any? = nil, asynchronous async: Bool = false, ddlog: DDLog = DDLog.sharedInstance) {
  if condition() == false {
    DDLogError(message, level: level, context: context, file: file, function: function, line: line, tag: tag, asynchronous: async, ddlog: ddlog)
    Swift.assertionFailure(message, file: file, line: line)
  }
}

public func DDAssertionFailure(_ message: @autoclosure () -> String = "", level: DDLogLevel = defaultDebugLevel, context: Int = 0, file: StaticString = #file, function: StaticString = #function, line: UInt = #line, tag: Any? = nil, asynchronous async: Bool = false, ddlog: DDLog = DDLog.sharedInstance) {
  DDLogError(message, level: level, context: context, file: file, function: function, line: line, tag: tag, asynchronous: async, ddlog: ddlog)
  Swift.assertionFailure(message, file: file, line: line)
}
```

_Edit: fixed the condition check in `DDAssert` (because I'm stupid) and added a default value for the `message` parameter so that you can call these without a message string._

diederich:
[...]

PR #934, opened 2018-05-19
DDAssert for Swift

ole:

* [ ] I have added the required tests to prove the fix/feature I am adding _(see discussion below)_
* [ ] I have updated the documentation (if necessary) _(not sure if these functions should be mentioned in the documentation)_

### Pull Request Description

This PR contains the implementation for #918: two new Swift functions, `DDAssert` and `DDAssertionFailure`. They mimic the stdlib's [`assert`](https://developer.apple.com/documentation/swift/1541112-assert) and [`assertionFailure`](https://developer.apple.com/documentation/swift/1539616-assertionfailure), but they also log a message (using `DDLogError`) if the assertions fails. (The consequence is that, unlike `assert`, the assertion condition for `DDAssert` is always evaluated, even when assertions are disabled.)

cc @bpoplauschi @diederich 

#### Implementation notes

* I chose to create a new file, `DDAssert.swift`, for these function. If you prefer, we could also include them in `CocoaLumberjack.swift`, of course. I found it clearer to separate the two because the new functions aren't plain Obj-C wrappers.

* I added the new file to the podspec, but I didn't change anything else (version number etc.) in the podspec because I'm not familiar with your release process. @bpoplauschi Please advise if I should change anything else.

#### Missing tests

This PR doesn't come with any new unit tests. The main reason is that testing these functions is inherently difficult (or impossible). The only "simple" test I can come up with is a test that `DDAssert` does _not_ log anything when the assertion passes.

Testing for positive logs when assertions fail would either (a) require us to build the test targets with assertions disabled, or (b) add another flag parameter to the functions that would only be used by the tests, e.g.

```swift
func DDAssert(_ condition: ..., ..., omitCallToAssert: Bool = false)
```

**In my opinion option (a) is problematic, but option (b) isn’t too bad. I think it would be a nice compromise. Option (c) would be to leave out any tests. What do you think?**

Less importantly, we'd have to set up the test targets for tests written in Swift. This would entail things like:

* Upping the deployment target of the test target to iOS 8/macOS 10.10 (the minimum version supported by Swift). Is that a problem?
* Implementing a mock `DDLogger` implementation in Swift (OCMock can't be used in test written in Swift).

I have a partial implementation of this on a branch: [ddassert-swift-tests](https://github.com/ole/CocoaLumberjack/tree/ddassert-swift-tests).

#### Decision Required

Should we:

* merge the PR as is (without tests)?
* add a `omitCallToAssert` flag to the functions and write tests using the flag?
* do something else?

### DIFF
--- a/Classes/DDAssert.swift
+++ b/Classes/DDAssert.swift
@@ -1,0 +1,43 @@
+// Software License Agreement (BSD License)
+//
+// Copyright (c) 2014-2016, Deusty, LLC
+// All rights reserved.
+//
+// Redistribution and use of this software in source and binary forms,
+// with or without modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Neither the name of Deusty nor the names of its contributors may be used
+//   to endorse or promote products derived from this software without specific
+//   prior written permission of Deusty, LLC.
+
+/**
+ * Replacement for Swift's `assert` function that will output a log message even when assertions
+ * are disabled.
+ *
+ * - Parameters:
+ *   - condition: The condition to test. Unlike `Swift.assert`, `condition` is always evaluated,
+ *     even when assertions are disabled.
+ *   - message: A string to log (using `DDLogError`) if `condition` evaluates to `false`.
+ *     The default is an empty string.
+ */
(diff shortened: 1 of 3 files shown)


ole:
@bpoplauschi Thanks for the invitation. I applaud your (and others') efforts to keep this project running, but I'd rather decline.

bpoplauschi:
@ole sure, no problem. Feel free to make other contributions if and when appropriate.
Take care.

PR #950, opened 2018-09-28
Update to Xcode 10 and Swift 4.2 complier

sushichop:

### Pull Request Description

Update to Xcode 10 and Swift 4.2 complier.
I set default SWIFT_VERSION to 4.2. 

In addition, this time I made CocoaLumberjack work with SWIFT_VERSION=4.2, SWIFT_VERSION=4.0, or SWIFT_VERSION=3.0 because some people still use Swift 3.
But if we don't need to support multi Swift versions, I'll change code to support only Swift 4.2. 

What do you think?

FYI
According to https://docs.swift.org/swift-book/GuidedTour/Compatibility.html,
Swift 4.2 complier with SWIFT_VERSION=3.0 means Swift 3.4.)

Thanks in advance.


bpoplauschi:
Thanks for taking an interest in this @sushichop. I have started a branch for the Xcode10 support, but is not ready yet - diverged a bit into adding xcconfig files and merging some of the targets, but that is another story :)

For now, I'm not sure we need to go back with the Swift compatibility, as the current version 3.4.2 requires Swift 4.0. I think it's enough to bump that to 4.2. Also keeping 2 `pod lib lint` phases just makes the builds even longer, I don't think we need them.

sushichop:
@bpoplauschi 
Thanks for your response.

> For now, I'm not sure we need to go back with the Swift compatibility, as the current version 3.4.2 requires Swift 4.0. I think it's enough to bump that to 4.2. Also keeping 2 `pod lib lint` phases just makes the builds even longer, I don't think we need them.

I agree with you.
Then, I changed code in order to support only Swift 4.2.

sushichop:
@bpoplauschi 
Thanks. I have merged!

bpoplauschi:
Good, I was waiting for the CI build to finish

Issue #774, opened 2016-09-20
Logging from Swift leaks Strings


c0diq:
### New Issue Checklist
- [X] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
- [X] I have read the [Documentation](https://github.com/CocoaLumberjack/CocoaLumberjack#documentation) or [CocoaDocs](cocoadocs.org/docsets/CocoaLumberjack)
- [X] I have searched for a similar issue in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/issues) and found none
### Issue Info

 Platform Name           | ios
 Platform Version        | 9.0
 CocoaLumberjack Version | 2.4.0
 Integration Method      | manually
 Xcode Version           | Xcode 8.0
 Repro rate              | all the time (100%)
### Issue Description and Steps

When logging from Swift using CocoaLumberjackSwift framework, the strings are not removed from the binary when the level indicates it should.
This is because Swift compiler cannot optimize out code like it does with Objective-C when using static const log level inside macros.
As such, the string "Hello World" can be found in the binary when invoking

```
DDLogDebug(message: "Hello World", level: .DDLogLevelInfo)
```

Expected behavior is that it should be compiled out.

Interestingly, this code properly removes the string from the binary:

```
if DDLogLevel.Info.rawValue & DDLogFlag.Debug.rawValue != 0 {
      let logMessage = DDLogMessage(message: "Hello World", level: .Info, flag: .Debug, context: 0, file: "foo", function: "bar", line: #line, tag: nil, options: [.CopyFile, .CopyFunction], timestamp: nil)
      DDLog.sharedInstance().log(true, message: logMessage)
}
```

diederich:
Is this for Debug or Release builds? What are the optimisation settings?

c0diq:
Release builds. Swift Fast Single-File Optimization [-O].

bpoplauschi:
This is beyond my knowledge. If anyone knows how to fix this, please let us know.
@rivera-ernesto @MaddTheSane

ffried:
I've just reproduced this with Xcode 10 using Swift 4.2 (with an iOS 12 target).

I also added inlining attributes to the Swift DDLog... functions. And in the binary these functions actually were inlined. However the strings still were in the binary, too.

[...]

PR #952, opened 2018-10-05
Use static const for default log level so Swift can strip strings

ffried:

This merge request fixes / refers to the following issues: #774 

### Pull Request Description

As described in #774, the previous `defaultDebugLevel` was a normal global variable, that could change at any time. Therefor the compiler did not strip strings that were logged with an inaccessible level.
This changes this by adding `@inlinable` to the Swift log functions (so the compiler is allowed to inline the code) and using a `static const` definition for the default log level. This definition now cannot be changed in Swift code, but needs to be defined either via the `GCC_PREPROCESSOR_DEFINITIONS` build setting, or using a `#define` in a bridging header of a Swift target.

### DIFF
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -46,9 +46,36 @@ extension DDLogFlag {
     }
 }
 
+/// The log level that can dynamically limit log messages (vs. the static DDDefaultLogLevel). This log level will only be checked, if the message passes the `DDDefaultLogLevel`.
+public var dynamicLogLevel = DDLogLevel.all
+
+/// Resets the `dynamicLogLevel` to `.all`.
+/// - SeeAlso: `dynamicLogLevel`
+@inlinable
+public func resetDynamicLogLevel() {
+    dynamicLogLevel = .all
+}
+
+
+@available(*, deprecated, message: "Please use dynamicLogLevel", renamed: "dynamicLogLevel")
+public var defaultDebugLevel: DDLogLevel {
+    get {
+        return dynamicLogLevel
+    }
+    set {
+        dynamicLogLevel = newValue
+    }
+}
+
+@available(*, deprecated, message: "Please use resetDynamicLogLevel", renamed: "resetDynamicLogLevel")
+public func resetDefaultDebugLevel() {
+    resetDynamicLogLevel()
+}
+
 @inlinable
 public func _DDLogMessage(_ message: @autoclosure () -> String, level: DDLogLevel, flag: DDLogFlag, context: Int, file: StaticString, function: StaticString, line: UInt, tag: Any?, asynchronous: Bool, ddlog: DDLog) {
-    if level.rawValue & flag.rawValue != 0 {
+    // The `dynamicLogLevel` will always be checked here (instead of being passed in). We cannot "mix" it with the `DDDefaultLogLevel`, because otherwise the compiler won't strip strings that are not logged.
+    if level.rawValue & flag.rawValue != 0 && dynamicLogLevel.rawValue & flag.rawValue != 0 {
         // Tell the DDLogMessage constructor to copy the C strings that get passed to it.
         let logMessage = DDLogMessage(message: message(), level: level, flag: flag, context: context, file: String(describing: file), function: String(describing: function), line: line, tag: tag, options: [.copyFile, .copyFunction], timestamp: nil)
         ddlog.log(asynchronous: asynchronous, message: logMessage)
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -46,12 +46,7 @@ extension DDLogFlag {
     }
 }
 
-public var defaultDebugLevel = DDLogLevel.verbose
-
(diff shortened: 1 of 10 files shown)


rivera-ernesto:
Looking good. Still WIP? Feel free to make multiple PR's.

ffried:
Mainly still WIP since I need to update the docs (which I could do in another PR however).

But I think we need to be careful on how we integrate this. This will break existing Swift code and remove the possibility to change the default log level at runtime in Swift.
I don't know if this is something that is needed. If so, we could add another layer below the `static const` (which of course then wouldn't strip strings).
So from a release perspective, this should go at least into a new minor version. If we want to stick strictly to semantic versioning, it should even go into a major release.

ffried:
To elaborate a bit on how to re-add changing the log level at runtime:
I'd re-add the `defaultDebugLevel` (though probably rename it, since it's not only for _debug_) and add the bitmask-check *after* the bitmask-check of the `static const`. Meaning that if the `static const` is disabling logging for the level of a log message, all bets are off and it's not going to be logged (and in release builds also stripped from the binary). If the `static const` allows the logging of a given message, however, it'll then be checked against the global Swift `var`. And only if both checks succeed, it'll be passed to `DDLog`.
Of course, this also introduces more complexity (e.g. questions like "why is my verbose log message not logged, I've set `defaultDebugLevel` to `.all`?" which could happen because `DD_LOG_LEVEL` is set to `DDLogLevelWarning`). On the other hand if we also make the default `.all` for `defaultDebugLevel`, this should only happen if both `defaultDebugLevel` *and* `DD_LOG_LEVEL` have already been changed. And at that point I'd expect the developer to know that.

ffried:
So, to summarize: this one right here is ready to be merged (minus the docs which I'd add in another PR).
If we want to re-add changing the log level in Swift at runtime, I'd also create another PR for this.
This would then also make it possible to ship this in a minor release (since it does not entirely remove a feature).

sushichop:
@ffried 

Thanks for explaining in detail😁

> But I think we need to be careful on how we integrate this. This will break existing Swift code and remove the possibility to change the default log level at runtime in Swift.

I agree with you. I concerned about this when I looked at this PR.

> If we want to re-add changing the log level in Swift at runtime, I'd also create another PR for this.

Great!! IMO, we (and many developers) need to be able to change log level in Swift at runtime.
I think this PR can be merged if you create another PR for it.

Then, which should we merge this PR right now or wait for another PR? 
(Can you let us know which one you prefer?)

Thanks in advance.

ffried:
@sushichop Thanks for confirming my concerns. Wasn't sure if I'm the only one having them. 🙂 


> Then, which should we merge this PR right now or wait for another PR?
(Can you let us know which one you prefer?)

This depends a bit on how close the next release is. I think I should be able to create the follow-up PR within the next week or so. So if there's another release planned until then, better wait with merging this PR. Otherwise this may as well be merged.

ffried:
On second thought, maybe it's better if I extend this PR (instead of creating another one). Otherwise the follow-up PR will just partially revert what I deleted in this PR. Especially since I basically need some of the now deleted Swift code (even if I rename the global var, I need to deprecate the old for a smooth transition).

sushichop:
@ffried 
OK. It seems that we just have better wait for this PR to be extended.

@rivera-ernesto @bpoplauschi 
What do think about this?

ffried:
@sushichop Done from my side. There is now a `dynamicLogLevel` with which a message can be "filtered" at runtime. It is, however, not used for setting the level of messages (like the `defaultDebugLevel` did before). The level of messages will always be set from `DDDefaultLogLevel`.

---
No idea why the codebeat check does not finish...

sushichop:
@ffried 
Thanks. LGTM😁

> No idea why the codebeat check does not finish...

I also have no idea...

@bpoplauschi @rivera-ernesto 
Do you know something about codebeat or codacy? 
And can you review this PR? Now, I think we are able to merge it.

rivera-ernesto:
> Do you know something about codebeat or codacy?

I added both. They just check coding style, common bugs, etc.
They just give us and contributors a second/third/fourth opinion on the PR.

> And can you review this PR? Now, I think we are able to merge it.

I tend to merge often. We have many users and if something's missing or wrong a new PR will come right after.

ffried:
Thanks! I'll create another PR with the docs update ASAP.

sushichop:
@rivera-ernesto 
All right. Thanks😁

bpoplauschi:
Very good contribution @ffried 
The only open question is: do we need a major release (i.e. 4.0.0) for this feature? I think it changes the library's API, but I'm not sure.

ffried:
@bpoplauschi now we don’t need a major release anymore. We would have with my first approach where I completely removed the possibility to change the log level in Swift at runtime. But since I added it back with a new variable name and properly deprecated the old one, this can also be released in a minor release now.
It still slightly changes the API now, but in a backwards compatible manner.

bpoplauschi:
@ffried perfect, I missed the adding back part :)
Then probably a minor release (3.5.0)

Issue #954, opened 2018-10-09
[DISCUSSION] Simplify CocoaLumberjack


sushichop:
### New Issue Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](https://github.com/CocoaLumberjack/CocoaLumberjack#documentation) or [CocoaDocs](cocoadocs.org/docsets/CocoaLumberjack)
* [x] I have searched for a similar issue in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/issues) and found none

IMHO, the structure of CocoaLumberjack is becoming complicated.
Then, I want to simplify CocoaLumberjack in order to maintain it flexibly and easily for the future.

**2 necessary things to simplify**
1)  I want to change minimum os versions to iOS 8.0, macOS 10.10, tvOS 9.0, and watchOS 2.0. 
These values should be low enough for many developers and, furthermore, these are the same as the minimum versions of most libraries.

2) I want to limit podspecs to only 2 specs. default(Objective-C) subspec and swift subspec.
I think most developers use only these 2 specs. Of course, I'll add the old subspecs(extension and cli subspecs) to the new default and swift subspec.

What do you think about this?
If you agree, I'll be glad to create a PR for it.

---
**Background**

After finishing the above, I want to introduce universal(multi platform) framework instead of each targeted platform framework.
 - CocoaLumberjack- {iOS, macOS, tvOS, watchOS} (4 frameworks) -> CocoaLumberjack (1 framework)
 - CocoaLumberjackSwift- {iOS, macOS, tvOS, watchOS} (4 frameworks) -> CocoaLumberjackSwift (1 framework)

I also think the simplicity as above is necessary as a first step to support SwiftPM in the near future.

Thanks.

ffried:
> @ffried the idea with xcconfig files was to simplify all our projects (since we have a lot in this repo) and to have a central place with these settings. I have used this approach in other projects and it reduces the conflicts, basically the xcodeproj will keep files and targets and the Xcode settings are in the xcconfig files.

[...]

PR #959, opened 2018-10-15
Merged framework targets + using xcconfig + deployment target iOS 8 and Mac OS 10.10

bpoplauschi:

### Pull Request Description

- Xcode 10 (already supported)
- merged the framework targets into multi-platform ones
- using xcconfig for most of the configurations (Lumberjack.xcodeproj + Tests.xcodeproj + Demos) - only adding setting in xcodeproj where they are not valid for all the other targets
- set deployment target to iOS8 and Mac OS 10.10 - also cleaned up some branched code for the earlier versions
- updated Travis scripts


lolgear:
@bpoplauschi could you also include separation of framework target and integration target?

https://github.com/CocoaLumberjack/CocoaLumberjack/issues/954#issuecomment-429794812

bpoplauschi:
@lolgear I will look into this

bpoplauschi:
@CocoaLumberjack/collaborators please review this PR - I would like to merge it sooner rather than later so @sushichop can work on top of it.

bpoplauschi:
@lolgear I haven't done the splitting of targets into 2 projects. I thought that will just make this PR even longer to complete. Perhaps we/you can do that in a separate one after this one is merged.

lolgear:
@bpoplauschi could you create milestone for "refactoring"?

sushichop:
@bpoplauschi 
Great Job! I gladly merge this PR

bpoplauschi:
Thanks @sushichop 

@lolgear what do you mean by a milestone for refactoring? Like a major release 4.x where we can refactor? We can start with a `4.x` branch and work there. What ideas of refactoring do you have, cause we should talk in advance. For moving the app integration targets to a separate project we don't need a major refactoring.

bpoplauschi:
@diederich 

> I might break for some people, but we probably won't find out without merging.

Do you mean for our users? If so, what do you think it will break? Other than removing support for iOS 6 and 7 and Mac OS X 10.8 and 10.9? :)

> I like xcconfig files. If you want something that has more comments, maybe you can copy from here: https://github.com/jspahrsummers/xcconfigs/blob/master/Base/Common.xcconfig

Thanks for the suggestion. I think it's a lot clearer if they have comments, so I added them: e97da34 - committed directly since it only adds comments.

lolgear:
@bpoplauschi

Refactoring or 3.5.0, it doesn't matter. New milestone is needed. "Refactoring-at-3.5.0" would be better, because all intensions in 3.5.0 are directed to refactoring and cleanup. ( and fix bugs, btw )

bpoplauschi:
@lolgear I added the 3.5.0 milestone - we only have a few remaining things to do before we can release it.

PR #961, opened 2018-10-16
integration: app targets have been separated from framework.

lolgear:
### New Pull Request Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](http://cocoadocs.org/docsets/CocoaLumberjack/)
* [x] I have searched for a similar pull request in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/pulls) and found none

* [x] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)
* [x] I have added the required tests to prove the fix/feature I am adding
* [x] I have updated the documentation (if necessary)
* [x] I have run the tests and they pass
* [x] I have run the lint and it passes (`pod lib lint`)

This merge request fixes / refers to the following issues: ...

### Pull Request Description

App targets have been separated from framework.

* [] Check watchOS and iOS Swift integration app targets. ( CocoaLumberjackSwift won't compile ).
* [] Remove old integration targets.
* [] Update travis configuration file.

PR #965, opened 2018-10-18
Cleanups

bpoplauschi:

This merge request fixes / refers to the following issues: #961 

### Pull Request Description

- added `Integration.xcodeproj` to the workspace
- fixed `macOSSwiftIntegration` build running issue (the `AppDelegate.swift` was never called - replaced with `main.swift`)
- removed `xcconfig` files from copy build phase
- continued the Xcode settings cleanup - moved most of them to `xcconfig`
- set Integration Apps `WATCHOS_DEPLOYMENT_TARGET` to `5.0` because of some APIs used
- fixed some app ids that were incorrect (`org.example.*` -> `com.deusty.*`)
- the `iOS` and `tvOS` targets `AppDelegate.swift` was not creating the `ViewController` instance. Fixed
- set explicit dependencies between the Integration targets and their corresponding `CocoaLumberjack` dependencies
- cleanup old remaining `AppIcon`
- `watchOSSwiftIntegration` now builds and runs (uses Debug config)
- `PREPROCESSOR_MACROS` is actually not used by Xcode, so the `DD_LOG_LEVEL=DDLogLevel*` setting was not working. Changed to `GCC_PREPROCESSOR_DEFINITIONS` and now it works. Set `DDLogLevelWarning` for release and `DDLogLevelAll` for debug builds - only applies to `Integration` targets


ffried:
I‘ve added [WIP] to the title so Danger picks it up correctly. 🙂

bpoplauschi:
@ffried forgot about that :)

LumberjackPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">PR is classed as Work in Progress</td>
    </tr>
  </tbody>
</table>

</p>

ffried:
Just in time. 😃

lolgear:
@bpoplauschi 
Could it be merged?
It freezes further development :/

bpoplauschi:
@lolgear not yet - that's why (actually @ffried) marked it as WIP.
The targets are building but I had some issues running the watchOS demo on simulator, so I want to look into that.

Integration/macOSSwiftIntegration/main.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  main.swift
+//  macOSSwiftIntegration
+//
+//  Created by Bogdan Poplauschi on 18/10/2018.
+//  Copyright © 2018 CocoaLumberjack. All rights reserved.
+//
```
ffried: Don't forget to copy the new copyright header in here, too.

Integration/Sources/ViewController.swift
```diff
@@ -29,8 +29,8 @@ class ViewController: UIViewController {
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         let formatter = Formatter()
-        DDTTYLogger.sharedInstance.logFormatter = formatter
-        DDLog.add(DDTTYLogger.sharedInstance)
+        DDOSLogger.sharedInstance.logFormatter = formatter
+        DDLog.add(DDOSLogger.sharedInstance)
```
ffried: I might have accidentally fixed that already in #968...

Integration/macOSSwiftIntegration/main.swift
bpoplauschi: Yeap, good catch.

Integration/Sources/ViewController.swift
bpoplauschi: No problem :)

Integration/macOSSwiftIntegration/main.swift
bpoplauschi: @ffried Hmm. We should add a check into Danger, that way we don't get into this spot again.

ffried: For new files, this shouldn’t happen again now. I’ve added the templates to the Xcode project files. so this should only be a temporary problem now.

bpoplauschi: Yeah but people always remove templates :) same as for the GitHub issue template.

Here's what I come up with:
```ruby
# Check Copyright - first 7 lines only
expected_copyright_lines = [
"// Software License Agreement (BSD License)",
"//",
"// Copyright (c) 2010-2018, Deusty, LLC",
"// All rights reserved.",
"//",
"// Redistribution and use of this software in source and binary forms,",
"// with or without modification, are permitted provided that the following conditions are met:",
"//",
"// * Redistributions of source code must retain the above copyright notice,",
"//   this list of conditions and the following disclaimer.",
"//",
"// * Neither the name of Deusty nor the names of its contributors may be used",
"//   to endorse or promote products derived from this software without specific",
"//   prior written permission of Deusty, LLC."
]

sourcefiles_to_check = (git.modified_files + git.added_files).uniq
invalid_copyright = false
sourcefiles_to_check.each do |sourcefile|
  fileExtension = File.extname(sourcefile)
  next unless (fileExtension.include?(".swift") or fileExtension.include?(".h") or fileExtension.include?(".m"))
  next unless File.file?(sourcefile)
  File.readlines(sourcefile)[0..13].each_with_index do |line, index|
    if !line.include?(expected_copyright_lines[index])
      invalid_copyright = true
      markdown("Invalid copyright: " + sourcefile)
      break
    end
  end
end
if invalid_copyright === true
  warn("Copyright is not valid. See our default copyright in all of our files")
end
```

bpoplauschi:
Current error when running the `watchOSSwiftIntegration` app: 
```
dyld: Library not loaded: @rpath/libswiftCoreImage.dylib
  Referenced from: /Users/bogdan.poplauschi/Library/Developer/Xcode/DerivedData/Lumberjack-dywkskrjusgjjjavbrafbuzvazky/Build/Products/Release-iphonesimulator/CocoaLumberjackSwift.framework/CocoaLumberjackSwift
  Reason: image not found
```

lolgear:
@bpoplauschi 
Does "Always copy swift core libraries" option set to YES?

Integration/macOSSwiftIntegration/main.swift
bpoplauschi: :) I was initially checking for the default 7 lines Copyright, but then I thought we can actually check for the one we need.
I can also add a different Copyright for Demos.
Yeap, 13 is equal to the number of `expected_copyright_lines`. I thought I could check it directly like if the beginning of the file is x. Not a final version definitely.

ffried: For Demos, we need to check the following pattern:

```
//
//  $FILENAME
//  $TARGET
//
//  CococaLumberjack Demos
//
```

While `$FILENAME` is easy (just the name of the file), `$TARGET` is a little more complicated. For now I'd just suggest we ignore it. Especially since there's not always a unique target, in which case `$TARGET` might be chosen differently. But if the rest matches, I'm fine with it.

bpoplauschi:
@sushichop I didn't add any `ComplicationController`, but just made explicit the dependency to `CocoaLumberjack.swift`, fixed the bundle ids (from `Info.plist) from `org.example.*` to `com.deusty.*` and  set `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` to `YES` for both the `watchos` and `watchos extension` targets - it builds and runs fine - check this branch/pr

Configs/App-Debug.xcconfig
```diff
@@ -3,3 +3,9 @@
 
 #include "Module-Debug.xcconfig"
 #include "App-Shared.xcconfig"
+
+// Space-separated list of preprocessor macros of the form `foo` or `foo=bar`
+GCC_PREPROCESSOR_DEFINITIONS = $(inherited) DEBUG=1 DD_LOG_LEVEL=DDLogLevelAll
```
bpoplauschi: just to make it clear:
`Module-Debug.xcconfig` sets `GCC_PREPROCESSOR_DEFINITIONS = $(inherited) DEBUG=1` but in this case we are shadowing that setting, so that's why I made it explicit here.

Integration/macOSSwiftIntegration/main.swift
bpoplauschi: Check out #972 - I also included a simple check for the Demos.

bpoplauschi:
@CocoaLumberjack/collaborators - this PR is now ready for final review/merge.

bpoplauschi:
@ffried Thanks. I think it's cleaner if I don't merge it. Can you?

bpoplauschi:
Thanks

PR #970, opened 2018-10-19
Cleanup of tests

lolgear:
### New Pull Request Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](http://cocoadocs.org/docsets/CocoaLumberjack/)
* [x] I have searched for a similar pull request in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/pulls) and found none

* [x] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)
* [x] I have added the required tests to prove the fix/feature I am adding
* [x] I have updated the documentation (if necessary)
* [x] I have run the tests and they pass
* [x] I have run the lint and it passes (`pod lib lint`)

This merge request fixes / refers to the following issues: ...

### Pull Request Description

...
CocoaPods are removed from Tests.

PR #971, opened 2018-10-22
Cleanup after #970

ffried:

This merge request fixes / refers to the following issues: #970 

### Pull Request Description

This does some smaller cleanup after #970.
- Mostly code style.
- Some settings were set in the build settings instead of being derived from the xcconfigs.
- Re-added some scheme settings (coverage gathering)
- Enabled some new scheme settings (target-based coverage gathering, parallel testing and random test order).

### DIFF
--- a/Tests/Tests.xcodeproj/project.pbxproj
+++ b/Tests/Tests.xcodeproj/project.pbxproj
@@ -199,7 +199,7 @@
 					};
 				};
 			};
-			buildConfigurationList = 432B53201AAE40EB00843E69 /* Build configuration list for PBXProject "CocoaLumberjack Tests" */;
+			buildConfigurationList = 432B53201AAE40EB00843E69 /* Build configuration list for PBXProject "Tests" */;
 			compatibilityVersion = "Xcode 3.2";
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
@@ -322,8 +322,6 @@
 				INFOPLIST_FILE = Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/../Frameworks";
 				SUPPORTED_PLATFORMS = macosx;
-				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
-				SWIFT_VERSION = 4.2;
 				TARGETED_DEVICE_FAMILY = "";
 			};
 			name = Debug;
@@ -340,7 +338,6 @@
 				INFOPLIST_FILE = Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/../Frameworks";
 				SUPPORTED_PLATFORMS = macosx;
-				SWIFT_VERSION = 4.2;
 				TARGETED_DEVICE_FAMILY = "";
 			};
 			name = Release;
@@ -368,7 +365,7 @@
 /* End XCBuildConfiguration section */
 
 /* Begin XCConfigurationList section */
-		432B53201AAE40EB00843E69 /* Build configuration list for PBXProject "CocoaLumberjack Tests" */ = {
+		432B53201AAE40EB00843E69 /* Build configuration list for PBXProject "Tests" */ = {
 			isa = XCConfigurationList;
 			buildConfigurations = (
 				432B532A1AAE40EB00843E69 /* Debug */,
--- a/Tests/Tests.xcodeproj/xcshareddata/xcschemes/OS X Tests.xcscheme
+++ b/Tests/Tests.xcodeproj/xcshareddata/xcschemes/OS X Tests.xcscheme
@@ -10,10 +10,37 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      codeCoverageEnabled = "YES"
+      onlyGenerateCoverageForSpecifiedTargets = "YES"
       shouldUseLaunchSchemeArgsEnv = "YES">
+      <CodeCoverageTargets>
+         <BuildableReference
+            BuildableIdentifier = "primary"
+            BlueprintIdentifier = "19FF46011B8B4CF400B43179"
+            BuildableName = "CocoaLumberjack.framework"
+            BlueprintName = "CocoaLumberjack"
+            ReferencedContainer = "container:../Lumberjack.xcodeproj">
+         </BuildableReference>
+         <BuildableReference
+            BuildableIdentifier = "primary"
+            BlueprintIdentifier = "18F3BFD61A81E06E00692297"
+            BuildableName = "libCocoaLumberjack.a"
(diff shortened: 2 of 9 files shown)


lolgear:
@ffried 
Could you also add .clang-format file in this PR?
It would be nice to have common coding style rules.

bpoplauschi:
> @ffried
> Could you also add .clang-format file in this PR?
> It would be nice to have common coding style rules.

Maybe in a dedicated PR :)

lolgear:
@bpoplauschi @ffried 

Should DDTTYLogger be update to OSLogger in tests?

ffried:
As long as we have `DDTTYLogger` (and `DDASLLogger` for that matter), we should have tests for both, actually. But in places where they play a secondary role (as in not being the system under test), they could be replaced by `DDOSLogger`, yes.

lolgear:
@bpoplauschi @ffried 

Also I would like to remove all brackets and replace them with dot syntax ( where it possible ).

Could it be done in tests?

```objective-c
    XCTAssertEqual([DDLog allLoggersWithLevel].count, 3);
    XCTAssertEqual([[[DDLog allLoggersWithLevel] firstObject] level], DDLogLevelAll);
    XCTAssertEqual([[DDLog allLoggersWithLevel][1] level], DDLogLevelDebug);
    XCTAssertEqual([[DDLog allLoggersWithLevel][2] level], DDLogLevelInfo);
```

```objective-c
    XCTAssertEqual(DDLog.allLoggersWithLevel.count, 3);
    XCTAssertEqual(DDLog.allLoggersWithLevel.firstObject.level, DDLogLevelAll);
    XCTAssertEqual(DDLog.allLoggersWithLevel[1].level, DDLogLevelDebug);
    XCTAssertEqual(DDLog.allLoggersWithLevel[2].level, DDLogLevelInfo);
```

I mean

```objective-c
- (void)cleanup {
    [DDLog removeAllLoggers];
    [DDLog addLogger:[DDTTYLogger sharedInstance]];
    [DDLog addLogger:self.logger];
    
    ddLogLevel = DDLogLevelVerbose;
    
    self.logs = @[];
    self.expectation = nil;
    self.noOfMessagesLogged = 0;
}
```

ffried:
> Also I would like to remove all brackets and replace them with dot syntax ( where it possible ).

I'd stick to the property-convention here. Meaning properties should be accessed in dot-notation. Methods (even with no parameters) should not.

lolgear:
@ffried 
`[DDLog allLoggersWithLevel]` - ok, method.

`[[array firstObject] level]` - methods or properties?

ffried:
Just check the declarations:

`- [DDLogMessage level]`:
`@property (readonly, nonatomic) DDLogLevel level;`

`- [NSArray firstObject]`:
`@property (nullable, nonatomic, readonly) ObjectType firstObject API_AVAILABLE(macos(10.6), ios(4.0), watchos(2.0), tvos(9.0));`

ffried:
Even `DDLog allLoggersWithLevel`:
`@property (class, nonatomic, copy, readonly) NSArray<DDLoggerInformation *> *allLoggersWithLevel;`

bpoplauschi:
> I'd stick to the property-convention here. Meaning properties should be accessed in dot-notation. Methods (even with no parameters) should not.

That is my preference too.

In this case
`@property (class, nonatomic, copy, readonly) NSArray<DDLoggerInformation *> *allLoggersWithLevel;`
and
`@property (nonatomic, readonly) DDLogLevel level;`
so
`DDLog.allLoggersWithLevel[0].level`

lolgear:
@bpoplauschi @ffried 

It is weird.

```objective-c
    __auto_type atomicCounter = [[DDAtomicCounter alloc] initWithDefaultValue:0];
    XCTAssertEqual([atomicCounter value], 0);
    XCTAssertEqual([atomicCounter increment], 1);
    XCTAssertEqual([atomicCounter value], 1);
    XCTAssertEqual([atomicCounter decrement], 0);
    XCTAssertEqual([atomicCounter value], 0);
```

Declaration:

```objective-c
- (int32_t)increment;
- (int32_t)decrement;
- (int32_t)value;
```

Methods:

```objective-c
- (int32_t)value {
    return _value;
}

#if DD_OSATOMIC_API_DEPRECATED
- (int32_t)increment {
    atomic_fetch_add_explicit(&_value, 1, memory_order_relaxed);
    return _value;
}

- (int32_t)decrement {
    atomic_fetch_sub_explicit(&_value, 1, memory_order_relaxed);
    return _value;
}
#else
- (int32_t)increment {
    return OSAtomicIncrement32(&_value);
}

- (int32_t)decrement {
    return OSAtomicDecrement32(&_value);
}
#endif
```

As we see `value` only returns variable.

However, it is declared as a method. But all loggers declared as property.

```
@property (class, nonatomic, copy, readonly) NSArray<DDLoggerInformation *> *allLoggersWithLevel;
```

And in implementation it is super-heavy task:

1. Accessing `shared instance`.
2. in `synced queue` in `autorelease pool`.
3. gather all loggers in `block` variable and return it.

ffried:
As for `value`: Should definitively be a property.

As for `allLoggersWithLevel`: It sure looks like it does a lot. But most of it is just "making access thread-safe", which properties are allowed to do. There isn't any heavy computation to create a `DDLoggerInformation` object, so this all just breaks down to a thread-safe iteration of an array.

bpoplauschi:
Super agree with @ffried

lolgear:
What about `increment` and `decrement`?

These methods have atomic `increment` and `decrement` correspondingly and return new `value`.

bpoplauschi:
From my pov `increment` and `decrement` cannot be properties :) they are definitely functions/methods

ffried:
Agree with @bpoplauschi: `increment` and `decrement` perform actions that change `value`. It's convenience, that they return the new value. They wouldn't have to.

ffried:
Properties (their getters) should be idempotent. So they should always return the same result (given that nothing else changes).

lolgear:
@ffried 
It is time for Bug Report! :)

ffried:
@lolgear: For a dead language? 😜

lolgear:
@ffried 
Not sure that it is true 🍰 

I would like to wake up in the morning and read news: "Objective-C 3 is coming out." 😄

ffried:
@lolgear It's definitively a controversial topic. And above's comment is my personal opinion!
We replaced all our ObjC code with Swift and never regretted anything.
Bug again: personal opinion and personal experiences. Your mileage may vary. 😉

bpoplauschi:
@lolgear 

> I would like to replace ObjC but I am aware of user experience.
> Simple app ~50mb - not ok.

This will change very soon. With Swift 5 we will finally have ABI stability, which means we will no longer need to include all the `lib*swift*.dylib` within each app. That will reduce the size big time.

ffried:
@lolgear: If size matters, then maybe. But to be honest in my experience most people don't care if they download 5MB or 50MB. Twitter.app is 100MB and I've just read through the negative reviews and haven't found one about the size of the app.
I agree that there are probably regions on this earth where app size matters more (due to internet speed and volume limits). But this is only a question of time and won't be solved by everyone writing small apps.

Also, App Thinning does a lot here. So will ABI stability with being able to ship / update the stdlib and overlays separately. Even APFS does some good (by only storing duplicate files once) in that matter.

PR #984, opened 2018-11-08
OSLog inline macros pitch.

lolgear:
### WARNING
It is a pitch.

### Pull Request Description

I would like to suggest macros for direct access to os logger.

Pros:

- automatic capturing line number, file etc.
- frameworks filtering ( by providing bundle identifier. )
- native and fast

### DIFF
--- a/Classes/DDLogMacros.h
+++ b/Classes/DDLogMacros.h
@@ -99,3 +99,23 @@
 #define DDLogInfoToDDLog(ddlog, frmt, ...)    LOG_MAYBE_TO_DDLOG(ddlog, LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagInfo,    0, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
 #define DDLogDebugToDDLog(ddlog, frmt, ...)   LOG_MAYBE_TO_DDLOG(ddlog, LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagDebug,   0, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
 #define DDLogVerboseToDDLog(ddlog, frmt, ...) LOG_MAYBE_TO_DDLOG(ddlog, LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagVerbose, 0, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
+
+// OSLog macros.
+
+#import <os/log.h>
+/*
+ OS_LOG_TYPE_DEFAULT = 0x00,
+ OS_LOG_TYPE_INFO    = 0x01,
+ OS_LOG_TYPE_DEBUG   = 0x02,
+ OS_LOG_TYPE_ERROR   = 0x10,
+ OS_LOG_TYPE_FAULT   = 0x11,
+ */
+
+#define DDOSLogCStringFromString(string) ([string cStringUsingEncoding:NSUTF8StringEncoding])
+#define DDOSLogGetBundleIdentifier ([NSBundle bundleForClass:self.class].bundleIdentifier) // convert to (char *)
+#define DDOSLogWithBundleIndetifier(subsystem, category) os_log_create(subsystem, category) // both parameters are (char *)
+
+#define DDOSLog(subsystem, category, level, frmt, ...) os_log_with_type(DDOSLogWithBundleIndetifier(subsystem, category), OS_LOG_TYPE_DEBUG, format, ##__VA_ARGS__)
+
+#define DDOSLogError(subsystem, category, frmt, ...) DDOSLog(subsystem, category, OS_LOG_TYPE_ERROR, frmt, ##__VA_ARGS__)
+#define DDOSLogError(category, frmt, ...) DDOSLogError(DDOSLogGetBundleIdentifier, category, frmt, ##__VA_ARGS__)
--- a/Classes/DDLogMacros.h
+++ b/Classes/DDLogMacros.h
@@ -102,6 +102,7 @@
 
 // OSLog macros.
 
+#if __has_include(<os/log.h>)
 #import <os/log.h>
 /*
  OS_LOG_TYPE_DEFAULT = 0x00,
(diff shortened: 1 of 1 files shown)


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lolgear:
Possible roadmap and goals.

1. Add separate macros functions to call os_logger directly.
2. Add flag to determine if user wants to use os_logger directly or via CocoaLumberjack.
3. Add mapping `DDLogError` and friends to new macros.
4. Add handy accessor which returns custom os_logger instance.

Classes/DDLogMacros.h
```diff
@@ -99,3 +99,23 @@
...
+/*
+ OS_LOG_TYPE_DEFAULT = 0x00,
+ OS_LOG_TYPE_INFO    = 0x01,
+ OS_LOG_TYPE_DEBUG   = 0x02,
+ OS_LOG_TYPE_ERROR   = 0x10,
+ OS_LOG_TYPE_FAULT   = 0x11,
+ */
+
+#define DDOSLogCStringFromString(string) ([string cStringUsingEncoding:NSUTF8StringEncoding])
+#define DDOSLogGetBundleIdentifier ([NSBundle bundleForClass:self.class].bundleIdentifier) // convert to (char *)
```
ffried: This won't work if the macro is used in a C-function.

Classes/DDLogMacros.h
```diff
@@ -99,3 +99,23 @@
...
+#define DDOSLogWithBundleIndetifier(subsystem, category) os_log_create(subsystem, category) // both parameters are (char *)
```
ffried: This macro does not do what its name says, right?

Classes/DDLogMacros.h
```diff
@@ -99,3 +99,23 @@
...
+#import <os/log.h>
+/*
+ OS_LOG_TYPE_DEFAULT = 0x00,
+ OS_LOG_TYPE_INFO    = 0x01,
+ OS_LOG_TYPE_DEBUG   = 0x02,
+ OS_LOG_TYPE_ERROR   = 0x10,
+ OS_LOG_TYPE_FAULT   = 0x11,
+ */
+
+#define DDOSLogCStringFromString(string) ([string cStringUsingEncoding:NSUTF8StringEncoding])
```
ffried: Seems unused?

ffried:
I've added some comments. But generally I don't think this is a good idea. It circumvents DDLog completely. No formatters, no file logs, nothing. IMHO, if someone wants to use OSLog directly, they should do so and not use CocoaLumberjack.

lolgear:
@ffried 
`os/log` is the end of CocoaLumberjack. It has everything that CocoaLumberjack has and even more.

```
No formatters, no file logs, nothing
```

os_log has everything.
It automatically write to files, you can setup `os_log` to write to disk or to memory.
It has great format engine with auto-capturing file name, line. It also gives you subsystem filtering.

[os log apple wwdc](https://developer.apple.com/videos/play/wwdc2016/721/)

ffried:
@lolgear Is that so. Then how do you get the log files **inside** the application? E.g. in all our apps, we have a feedback form that can attach the logs to the generated feedback email. With os_log this is only possible by creating a sysdiagnose - which cannot be done by the application itself.

os_log is great, no doubt. I use it heavily (e.g. in frameworks that can't have a dependency to CocoaLumberjack). But it does **not** provide all the features of CocoaLumberjack (yet). There's no way to have custom formatters. Yes, os_log does some neat formatting, but you cannot have custom formatters. You can log to a file, but only if doing a sysdiagnose. Filename and line are captured by CocoaLumberjack, too. In fact, I'm missing the possibility to pass that on to os_log - which is IMHO a missing key part in the API making writing wrappers around it incredibly hard.
Yes, os_log has subsystem filtering. But again it is mostly outside of the control of the application that is writing logs.

ffried:
I mean, you're not completely wrong. While I previously missed CocoaLumberjack in projects that could not use it, with os_log it's not that bad anymore. It is indeed a very nice logging system. But it's not a complete replacement for CocoaLumberjack (yet).

lolgear:
@ffried 
About sysdiagnose.

Some system frameworks require profile to gather logs. Person must download it and initiate sysdiagnose.

For example, Bluetooth requires profile to gather system logs. And you must download this profile and deliver to your user. ( Only developer can access logs and profiles from developer.apple )

For most apps, of course, `sysdiagnose` is something unnatural. But `sysdiagnose` is a heart of Apple diagnostics and each user, who wrote to Apple, must trigger diagnostics to get context and logs of a problem.
It is culture of Apple, not an `unnatural` way of interacting with Apple ecosystem and Apple devices.

Back to Bluetooth.

It is what person needs. Apple eliminates one step of confusion. Now and then person will trigger sysdiagnose and for Apple it must download some profiles to access logs. Maybe this practice is the best for future apps.

lolgear:
@ffried 
For user "another way" or "another command" in phone.

For example, I had known how to turn off phone until friend told me about mechanic keys combinations: `lock screen button` and `home button`.
I don't know about this combination because "phone just works". But in critical situation it "requires" combination.  Thus, maybe, it should tell me about it or I should google it.

And if Apple make unified logging for whole ecosystem across all apps, why person needs different step to log events and gather them? Person can read on trusted Apple site that "sysdiagnose" will trigger logs gathering. And it is what person expects. This step does what it should.

lolgear:
@ffried 
I can't say anything here.
All support teams that I wrote to, ask me to add steps to reproduce bugs. iOS/macOS - it doesn't matter.
For macOS app support team asks me also to watch in Console.app and to reveal any errors in it.

Besides, custom formatters and loggers require "read" tools. Efforts to parse logs is something that CocoaLumberjack doesn't provide.

Ok, I see. I suggest radical solution which isn't fit in "Apple unified logging" concept for third-party apps.

I would like to eliminate a gap between native `os log` and CocoaLumberjack loggers. Maybe correct solution is to **extend** macros by adding os log invocations?

ffried:
@lolgear 
Yea, we also ask for repro-steps. But sometimes users can't really provide them, but our log files provide the necessary hint to the problem. And sometimes we don't even have to ask for repro steps, because the logs tell us already where the problem lies.

> I would like to eliminate a gap between native `os log` and CocoaLumberjack loggers. Maybe correct solution is to **extend** macros by adding os log invocations?

This sounds more like something that could fit within CocoaLumberjack. But again, I'm not sure if people are happy, that e.g. their custom log formatters are not used for the logs written to os log.
From the perspective of CocoaLumberjack, os_log is "just another logger" (as currently implemented with `OSLogger`). IMHO the first thing we should do is improve OSLogger by for example adding the possibility to define a default subsystem and category.
Or we could implement the possibility to map CocoaLumberjack's DDLogMessage.tag and context to subsystems and / or categories.
As for file paths and line numbers: like I said, it's sad that those can't be passed to os_log. But as with other loggers, the easiest way to get them is to have a `DDLogFormatter` that adds them to the formatted log string.

Like with ASL before, if someone really wants to leverage **all** features of the API, they'll have to use it directly. Abstractions like CocoaLumberjack define a common superset of features. If someone needs a feature that is stripped away by this superset, then they need to use the underlying API directly.
Similar to CoreData being an abstraction over certain data stores, like e.g. SQLite. And while SQLite is able to e.g. dynamically add a table, CoreData needs it defined in its model file. If you need to dynamically add a table, you need to use SQLite directly.

lolgear:
@bpoplauschi 
Yes, I am interested in it. I can create teams and even projects ( "new" button is present ), however, "new" button is hidden in repositories.

What is the main intention and purpose of this project ( and also its name )?

bpoplauschi:
@lolgear I'm not sure what the name and purpose should be :) it's your idea.
I can just support you to add the project.
Like I said, that would be totally different from `CocoaLumberjack` and perhaps adapted to people that wan't a simpler logging solution just on top of `os_log`

@CocoaLumberjack/collaborators what do you think?

lolgear:
@bpoplauschi 
I am playing with fairy tales titles and end up with
`OSWizard` or something which reflects [The Wonderful Wizard of Oz](https://en.wikipedia.org/wiki/The_Wonderful_Wizard_of_Oz).

Maybe `CocoaWizardOfOS`?

bpoplauschi:
@lolgear I like the idea. I usually chose more boring names 😄 but I like it when projects have cool names. I think `OSWizard` is a bit ambiguous. But I do like something like `CocoaWizardOfOS` or `WizardOfOSLog`. I see no comments from other people, so you can choose it :)

hhanesand:
Closing due to decision to move this feature to another project.

PR #996, opened 2018-12-04
Fix availability checks in DDLog.m. Also fix memory leak. Fixes #995

### LINKED ISSUE #995: Memory Leak for macOS
in `DDLog.m`, there are some code for macos:
```
- (id)init {
    self = [super init];
    
    if (self) {
        self._loggers = [[NSMutableArray alloc] initWithCapacity:4];
        
#if TARGET_OS_IOS
        NSString *notificationName = @"UIApplicationWillTerminateNotification";
#else
        NSString *notificationName = nil;
        
        // On Command Line Tool apps AppKit may not be avaliable
#ifdef NSAppKitVersionNumber10_0
        
        if (NSApp) {
            notificationName = @"NSApplicationWillTerminateNotification";
        }
        
#endif
        
        if (!notificationName) {
            // If there is no NSApp -> we are running Command Line Tool app.
            // In this case terminate notification wouldn't be fired, so we use workaround.
            
            atexit_b (^{
                [self applicationWillTerminate:nil];
            });
        }
        
#endif /* if TARGET_OS_IOS */
        
```

There is a workaround:
```
            atexit_b (^{
                [self applicationWillTerminate:nil];
            });
```

It will cause DDLog will be retained and never be released. Right?

ffried:

This merge request fixes / refers to the following issues: #995 

### Pull Request Description

The availability checks in DDLog.m in `init` don't work anymore, because `NSAppKitVersionNumber10_0` is no longer a define.
This fixes this by checking for CLI tools with `DD_CLI` (which is used in other places) and `__has_include`.
Furthermore, the memory leak of custom DDLog objects captured in `atexit_b`.
And I've switched to use the constants provided by the frameworks for the notification names.

### DIFF
--- a/Classes/DDLog.m
+++ b/Classes/DDLog.m
@@ -28,6 +28,9 @@
 #import <Availability.h>
 #if TARGET_OS_IOS
     #import <UIKit/UIDevice.h>
+    #import <UIKit/UIApplication.h>
+#elif !defined(DD_CLI) && __has_include(<AppKit/NSApplication.h>)
+    #import <AppKit/NSApplication.h>
 #endif
 
 
@@ -180,24 +183,23 @@ static NSUInteger _numProcessors;
         self._loggers = [[NSMutableArray alloc] initWithCapacity:4];
         
 #if TARGET_OS_IOS
-        NSString *notificationName = @"UIApplicationWillTerminateNotification";
+        NSString *notificationName = UIApplicationWillTerminateNotification;
 #else
         NSString *notificationName = nil;
-        
+
         // On Command Line Tool apps AppKit may not be avaliable
-#ifdef NSAppKitVersionNumber10_0
-        
+#if !defined(DD_CLI) && __has_include(<AppKit/NSApplication.h>)
         if (NSApp) {
-            notificationName = @"NSApplicationWillTerminateNotification";
+            notificationName = NSApplicationWillTerminateNotification;
         }
-        
 #endif
-        
+
         if (!notificationName) {
             // If there is no NSApp -> we are running Command Line Tool app.
             // In this case terminate notification wouldn't be fired, so we use workaround.
+            __weak __auto_type weakSelf = self;
             atexit_b (^{
-                [self applicationWillTerminate:nil];
+                [weakSelf applicationWillTerminate:nil];
             });
         }
         
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Add new queue label which will be hold by a manual created thread #932
 - DDFileLogger log message is overridden #924
 - Fix thread safety issue in DDFileLogger #986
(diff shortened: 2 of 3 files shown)


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Classes/DDLog.m
```diff
@@ -180,24 +183,23 @@ - (id)init {
...
+            notificationName = NSApplicationWillTerminateNotification;
         }
-        
 #endif
-        
+
         if (!notificationName) {
             // If there is no NSApp -> we are running Command Line Tool app.
             // In this case terminate notification wouldn't be fired, so we use workaround.
+            __weak __auto_type weakSelf = self;
```
lolgear: is `auto_type` allowed in framework?

ffried: why wouldn't it?

lolgear: It is rather new feature that was introduced in iOS 10 (as I remember?) and shipped with Xcode 8. Heh, I don't remember correct clang version.

Well, if so, what is a policy of backward compatibility for outdated clang versions which were introduced since deployment target version lower bound.

If framework supports versions down to iOS 8, should it also support Xcode version in which iOS 8 was introduced? ( but precisely clang and llvm versions that were shipped with Xcode 6)

rivera-ernesto: I think Apple won't accept binaries build with older Xcode versions. Maybe that should be the Xcode compatibility policy.

ffried: Yea, I'm with @rivera-ernesto on this one.
Also, I'm sure if someone runs into a problem with this, they'll open an issue... 😜

lolgear: I have nice conversation about this feature with guy from Twitter. 
He said that this feature has "huge" impact on compilation time, especially for their codebase ( 1,5kk LoC ).

In my biggest project ( 200k LoC ) this feature doesn't have huge impact or any impact at all.

But..
If you start to "break one window", you should be consistent and break all windows by adding `__auto_type` for all pointers.

And here you will say "it is another PR" :) 
Or not?

Whirlwind:
I am waiting this PR 👍

rivera-ernesto:
Let's merge it already. We can always PR fixes (if needed) later on.

Classes/DDLog.m
ffried: Yea it would be another PR. Adding `__auto_type` doesn't fix #995 😉 
Also, for me this is something that can be added whenever a particular line is touched. Going through all source files just to add `__auto_type` doesn't seem to be worth the time. 😉

PR #998, opened 2018-12-06
Added support for log file headers

tmsrtx:
Added property `initialFileContents` to `DDLogFileManagerDefault` to support headers in log files.
Override implementation of `initialFileContents` to insert header text at the beginning of every log file created by the `createNewLogFile:` method.

### Pull Request Description

We wish to include the version and build date of our application at the beginning of the log files generated by CocoaLumberjack. I have not been able to find any obvious way of doing this so I propose adding an `initialFileContents` property to `DDLogFileManagerDefault` which the  `createNewLogFile:` method then uses as an argument for the 'contents' parameter (instead of nil) when calling `createFileAtPath:contents:attributes:`

The intention is to achieve something which is possible in NLog (though NLog is a very different beast), using `LayoutWithHeaderAndFooter`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Add new queue label which will be hold by a manual created thread #932
 - DDFileLogger log message is overridden #924
 - Fix thread safety issue in DDFileLogger #986
+- Added `initialFileContents` property to `DDLogFileManagerDefault` to set log file header
 
 ## [3.4.2 - Xcode 9.3 on Apr 17th, 2018](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.4.2)
 - Update README.md #912
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@
 - DDFileLogger log message is overridden #924
 - Fix thread safety issue in DDFileLogger #986
 - Fix availability checks and memory leak #996
-- Added `initialFileContents` property to `DDLogFileManagerDefault` to set log file header
+- Added `logFileHeader` property to `DDLogFileManagerDefault` which you can override to set log file header
 
 ## [3.4.2 - Xcode 9.3 on Apr 17th, 2018](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.4.2)
 - Update README.md #912
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -68,7 +68,7 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
 - (void)deleteOldLogFiles;
 - (NSString *)defaultLogsDirectory;
 
-@property (readonly, copy) NSData *initialFileContentsData;
+@property (readonly, copy) NSData *logFileHeaderData;
 
 @end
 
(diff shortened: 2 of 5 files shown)


rivera-ernesto:
Looks good, but I would rename `initialFileContents` with something like `logFileHeader`.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - DDFileLogger log message is overridden #924
 - Fix thread safety issue in DDFileLogger #986
 - Fix availability checks and memory leak #996
+- Added `initialFileContents` property to `DDLogFileManagerDefault` to set log file header
```
rivera-ernesto: One last left here and we're good to merge.

tmsrtx: Oh, right. Forgot that one, thanks :-)
I'm also implementing some test cases so both OSSPolice and codecov will hopefully be happy.

Tests/Tests/DDLogFileManagerTests.m
```diff
@@ -0,0 +1,154 @@
...
+
+@implementation DDLogFileManagerDefaultTests
+
+- (void)setUp {
+    [super setUp];
+    [self setUpLogFileManager];
+}
+
+- (void)setUpLogFileManager {
+    NSString *logsDirectory = NSTemporaryDirectory();
```
lolgear: Please, use `__auto_type` instead.

Tests are creatures from the modern world.

tmsrtx: Could you please explain to me what the benefit of using `__auto_type` in this case is?
It uses up more characters, does not explicitly tell a reader what the type is and, I assume, the compiled code will be the same. I can understand wanting to use `__auto_type` to infer the type when the resulting type is complex but here it is very simple.
I don't write Objective-C on an everyday basis so I'm just being curious here :-)

lolgear: @tmsrtx 

It is not the only one place, I just choose random place.

You have more "simple type" example.

```objective-c
        unsigned long long fileSize = [[[NSFileManager defaultManager] attributesOfItemAtPath:filePath error:nil][NSFileSize] unsignedLongLongValue];
```

```objective-c
        __auto_type fileSize = [[[NSFileManager defaultManager] attributesOfItemAtPath:filePath error:nil][NSFileSize] unsignedLongLongValue];
```

And, of course, writing mantras like: `NSString *something = [[NSString alloc] init..]` is not the best style.

What about top-level functions and this example?

```objective-c
    NSString *logsDirectory = NSTemporaryDirectory();
```

It is already incorrect. 

```objective-c
NSTemporaryDirectory(); // return type nonnull NSString *
```

For example, if `nullability` can become a part of type ( or any type placeholder construction ), my advise will take zero efforts.

lolgear:
However, adding "Header" for file logger is really discouraged.

It is not the point for a file logger.
File logger can gather logs into a file. Ok, it is a content of a file.
Next, if you like to post-process file, you can ( before archiving or sending to your server ) append header, footer, any type of stars, whistles, bells etc. You can also add your best poem or quote of the day.

rivera-ernesto:
I see what you say @lolgear , but at the same time by default it will be empty. So it looks good to me.

Classes/DDFileLogger.m
```diff
@@ -427,9 +429,28 @@ - (NSString *)newLogFileName {
...
+
+- (NSData *)logFileHeaderData {
+    NSString *fileHeaderStr = [self logFileHeader];
+    
+    if (fileHeaderStr == nil || [fileHeaderStr length] == 0) {
+        return nil;
+    }
+    
+    // Ensure that we have a newline at the end of the string
+    fileHeaderStr = [NSString stringWithFormat:@"%@\n", fileHeaderStr];
```
hhanesand:
```suggestion
    unichar lastChar = [fileHeaderStr characterAtIndex:fileHeaderStr.length - 1];
    if (![[NSCharacterSet newlineCharacterSet] characterIsMember:lastChar]) {
        fileHeaderStr = [NSString stringWithFormat:@"%@\n", fileHeaderStr];
    }
```

PR #1001, opened 2018-12-16
core: loggers file logger with buffer has been introduced.

lolgear:

### Pull Request Description

File logger with buffer has been introduced. It is a sketch.

### DIFF
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -136,6 +136,10 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
     @try {
         [self removeObserver:self forKeyPath:NSStringFromSelector(@selector(maximumNumberOfLogFiles))];
         [self removeObserver:self forKeyPath:NSStringFromSelector(@selector(logFilesDiskQuota))];
+        __auto_type path = [NSTemporaryDirectory() pathComponents];
+        __auto_type directory = NSTemporaryDirectory();
+        __auto_type components = [NSTemporaryDirectory() pathComponents];
+        __auto_type extension = [NSTemporaryDirectory() pathExtension];
     } @catch (NSException *exception) {
     }
 }
@@ -601,19 +605,23 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
     return self;
 }
 
-- (void)dealloc {
+- (void)cleanup {
     [_currentLogFileHandle synchronizeFile];
     [_currentLogFileHandle closeFile];
-
+    
     if (_currentLogFileVnode) {
         dispatch_source_cancel(_currentLogFileVnode);
         _currentLogFileVnode = NULL;
     }
-
+    
     if (_rollingTimer) {
         dispatch_source_cancel(_rollingTimer);
         _rollingTimer = NULL;
     }
+}
+
+- (void)dealloc {
+    [self cleanup];
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -1013,28 +1021,13 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 static int exception_count = 0;
-- (void)logMessage:(DDLogMessage *)logMessage {
-    NSString *message = logMessage->_message;
-    BOOL isFormatted = NO;
-
-    if (_logFormatter) {
-        message = [_logFormatter formatLogMessage:logMessage];
-        isFormatted = message != logMessage->_message;
-    }
-
-    if (message) {
-        if ((!isFormatted || _automaticallyAppendNewlineForCustomFormatters) &&
-            (![message hasSuffix:@"\n"])) {
-            message = [message stringByAppendingString:@"\n"];
-        }
-
-        NSData *logData = [message dataUsingEncoding:NSUTF8StringEncoding];
-
+- (void)logData:(NSData *)data {
+    if (data != nil) {
         @try {
             [self willLogMessage];
-			
+            
             [[self currentLogFileHandle] seekToEndOfFile];
-            [[self currentLogFileHandle] writeData:logData];
+            [[self currentLogFileHandle] writeData:data];
 
             [self didLogMessage];
         } @catch (NSException *exception) {
@@ -1048,6 +1041,26 @@ static int exception_count = 0;
                 }
             }
         }
+    }
+}
+
+- (void)logMessage:(DDLogMessage *)logMessage {
+    NSString *message = logMessage->_message;
+    BOOL isFormatted = NO;
+
+    if (_logFormatter != nil) {
+        message = [_logFormatter formatLogMessage:logMessage];
+        isFormatted = message != logMessage->_message;
+    }
+
+    if (message != nil) {
+        if ((!isFormatted || _automaticallyAppendNewlineForCustomFormatters) &&
+            (![message hasSuffix:@"\n"])) {
+            message = [message stringByAppendingString:@"\n"];
+        }
+
+        NSData *logData = [message dataUsingEncoding:NSUTF8StringEncoding];
+        [self logData:logData];
     }
 }
 
@@ -1531,3 +1544,101 @@ BOOL doesAppRunInBackground() {
 }
 
 #endif
+
+static NSUInteger kMaximumBytesCountInBuffer = (1 << 10) * (1 << 10); // 1 MB.
+static NSUInteger kDefaultBytesCountInBuffer = (1 << 10);
+@interface DDFileLoggerWithBuffer : DDFileLogger
+@property (assign, nonatomic, readwrite) NSUInteger maximumBytesCountInBuffer;
+@end
+
+@interface DDFileLoggerWithBuffer() {
+    NSOutputStream *_bufferStream;
+    NSUInteger _bufferSize;
+}
+@end
+
+@interface DDFileLoggerWithBuffer (StreamManipulation)
+- (void)flushBuffer;
+- (void)dumpBufferToDisk;
+- (void)appendToBufferData:(NSData *)data;
+@end
+
+@implementation DDFileLoggerWithBuffer (StreamManipulation)
+- (void)flushBuffer {
+    // do something.
+    [_bufferStream close];
+    _bufferStream = nil;
+    _bufferSize = 0;
+}
+- (void)dumpBufferToDisk {
+    // do something.
+    __auto_type data = (NSData *)[_bufferStream propertyForKey:NSStreamDataWrittenToMemoryStreamKey];
+    [super logData:data];
+    [self flushBuffer];
+}
+- (void)appendToBufferData:(NSData *)data {
+    __auto_type length = data.length;
+    if (data.length != 0) {
+        if (_bufferStream == nil) {
+            _bufferStream = [[NSOutputStream alloc] initToMemory];
+            [_bufferStream open];
+            _bufferSize = 0;
+        }
+        const uint8_t *appendedData = calloc(length, sizeof(uint8_t));
+        [data getBytes:(void *)appendedData length:length];
+        if (appendedData != NULL) {
+            [_bufferStream write:appendedData maxLength:length];
+        }
+        if (appendedData != NULL) {
+            free((void *)appendedData);
+        }
+        _bufferSize += _bufferSize + length;
+    }
+}
+@end
+
+@implementation DDFileLoggerWithBuffer
+@synthesize maximumBytesCountInBuffer = _maximumBytesCountInBuffer;
+- (instancetype)init {
+    if (self = [super init]) {
+        self.maximumBytesCountInBuffer = kDefaultBytesCountInBuffer;
+    }
+    return self;
+}
+
+#pragma mark - Subclass
+- (void)logData:(NSData *)data {
+    [self appendToBufferData:data];
+}
+
+#pragma mark - Initialization
+- (void)cleanup {
+    [self dumpBufferToDisk];
+    [super cleanup];
+}
+@end
+
+@interface DDFileLoggerWithBuffer (Logging)
+- (BOOL)shouldDumpBufferToDisk;
+- (void)putDataIntoBufferOrToDisk:(NSData *)data;
+@end
+
+@implementation DDFileLoggerWithBuffer (Logging)
+- (BOOL)shouldDumpBufferToDisk {
+    return _bufferSize > _maximumBytesCountInBuffer;
+}
+- (void)putDataIntoBufferOrToDisk:(NSData *)data {
+    if ([self shouldDumpBufferToDisk]) {
+        [self dumpBufferToDisk];
+    }
+    else {
+        [self logData:data];
+    }
+}
+@end
+
+@implementation DDFileLogger (ClassCluster)
++ (instancetype)createLoggerWithBuffer {
+    return [DDFileLoggerWithBuffer new];
+}
+@end
--- a/Classes/DDFileLogger.h
+++ b/Classes/DDFileLogger.h
@@ -515,3 +515,7 @@ extern unsigned long long const kDDDefaultLogFilesDiskQuota;
 - (NSComparisonResult)reverseCompareByModificationDate:(DDLogFileInfo *)another;
 
 @end
+
+@interface DDFileLogger (ClassCluster)
++ (instancetype)createLoggerWithBuffer;
+@end


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Classes/DDFileLogger.m
```diff
@@ -136,6 +136,10 @@ - (void)dealloc {
     @try {
         [self removeObserver:self forKeyPath:NSStringFromSelector(@selector(maximumNumberOfLogFiles))];
         [self removeObserver:self forKeyPath:NSStringFromSelector(@selector(logFilesDiskQuota))];
+        __auto_type path = [NSTemporaryDirectory() pathComponents];
```
ffried: These four values are unused.

Classes/DDFileLogger.m
```diff
@@ -601,21 +605,25 @@ - (instancetype)initWithLogFileManager:(id <DDLogFileManager>)aLogFileManager {
     return self;
 }
 
-- (void)dealloc {
+- (void)cleanup {
     [_currentLogFileHandle synchronizeFile];
     [_currentLogFileHandle closeFile];
-
+
```
ffried: Unnecessary spaces added

Classes/DDFileLogger.m
```diff
@@ -601,21 +605,25 @@ - (instancetype)initWithLogFileManager:(id <DDLogFileManager>)aLogFileManager {
...
     [_currentLogFileHandle synchronizeFile];
     [_currentLogFileHandle closeFile];
-
+    
     if (_currentLogFileVnode) {
         dispatch_source_cancel(_currentLogFileVnode);
         _currentLogFileVnode = NULL;
     }
-
+
```
ffried: Same as above.

Classes/DDFileLogger.m
```diff
@@ -1531,3 +1544,101 @@ BOOL doesAppRunInBackground() {
...
+@end
+
+@implementation DDFileLoggerWithBuffer (StreamManipulation)
+- (void)flushBuffer {
+    // do something.
+    [_bufferStream close];
+    _bufferStream = nil;
+    _bufferSize = 0;
+}
+- (void)dumpBufferToDisk {
```
ffried: Missing newline before method

Classes/DDFileLogger.m
```diff
@@ -1531,3 +1544,101 @@ BOOL doesAppRunInBackground() {
...
+    // do something.
+    __auto_type data = (NSData *)[_bufferStream propertyForKey:NSStreamDataWrittenToMemoryStreamKey];
+    [super logData:data];
```
ffried: Why not `[self logData:data];`?

Classes/DDFileLogger.m
```diff
@@ -1531,3 +1544,101 @@ BOOL doesAppRunInBackground() {
...
+    [self flushBuffer];
+}
+- (void)appendToBufferData:(NSData *)data {
```
ffried: Missing newline before method

Classes/DDFileLogger.m
```diff
@@ -1531,3 +1544,101 @@ BOOL doesAppRunInBackground() {
...
+@interface DDFileLoggerWithBuffer (Logging)
+- (BOOL)shouldDumpBufferToDisk;
+- (void)putDataIntoBufferOrToDisk:(NSData *)data;
+@end
+
+@implementation DDFileLoggerWithBuffer (Logging)
+- (BOOL)shouldDumpBufferToDisk {
+    return _bufferSize > _maximumBytesCountInBuffer;
+}
+- (void)putDataIntoBufferOrToDisk:(NSData *)data {
```
ffried: Missing newline before method

Classes/DDFileLogger.m
lolgear: @ffried 
I want to patch `logData` to decide which channel ( buffer or disk ) are preferable at this moment.
Here I have a mistake.

`logData` should be:

```objective-c
- (void)logData:(NSData *)data {
    [self putDataIntoBufferOrToDisk:data];
}
```

And `[super logData:data]` will do what it intended to: it put data to disk.

Classes/DDFileLogger.m
lolgear: @ffried 
Oh! It was a mistake. I just use it to show "the blessing of __auto_type" in different PR.

Classes/DDFileLogger.m
lolgear: @ffried 
Is there any .clang_format file in the repository?
It would be nice to have it to prevent further issues in style.

ffried: I've started one, but could not get everything to be styled correctly. Also, I'm not a big fan of "hard styling rules". There are a few basics, but e.g. I don't care if the implementation of a method is inline with its declaration as long as it's readable (e.g. `- (BOOL)shouldDoWork { return NO; }`)

My general advice here: Look at the code around you and style accordingly. 😉

Classes/DDFileLogger.m
ffried: Still it's kind of a "misuse" of `super`.
It's not clear (at least to me) when reading this code, that the super call effectively means "write data to disk".
But I see your intention. How about something like this?
```objc
- (void)logData:(NSData *)data {
    if ([self shouldDumpBufferToDisk]) {
        __auto_type bufferedData = (NSData *)[_bufferStream propertyForKey:NSStreamDataWrittenToMemoryStreamKey];
        [super logData:bufferedData]; // Write data to disk
        [self flushBuffer];
    }
    [self appendToBufferData:data];
}
```

`- (void)dumpBufferToDisk` and `- (void)putDataIntoBufferOrToDisk:(NSData *)data` would be removed completely and the `super` call would be inside the correct method.

Also, we only have the line `[self appendToBufferData:data]` once (instead of once in `if` and once in `else`)

Classes/DDFileLogger.m
```diff
@@ -1531,3 +1540,114 @@ BOOL doesAppRunInBackground() {
...
+}
+
+- (void)dumpBufferToDisk {
+    // do something.
+    __auto_type data = (NSData *)[_bufferStream propertyForKey:NSStreamDataWrittenToMemoryStreamKey];
+    [super logData:data];
+    [self flushBuffer];
+}
+
+- (void)appendToBufferData:(NSData *)data {
```
ffried: This should probably better be named `- (void)appendToBuffer:(NSData *)data`
```suggestion
- (void)appendToBuffer:(NSData *)data {
```

Classes/DDFileLogger.m
```diff
@@ -1531,3 +1540,114 @@ BOOL doesAppRunInBackground() {
...
+
+#pragma mark - Initialization
+- (void)cleanup {
+    [self dumpBufferToDisk];
+    [super cleanup];
+}
+
+@end
+
+@interface DDFileLoggerWithBuffer (Logging)
```
ffried: If we safe us the extra `@interface` and just use `#pragma mark - Logging`, the build won't fail because we use `putDataIntoBufferOrToDisk:` in `logData:` before it's declared 😉

lolgear:
@ffried 
I am even not sure that this should be done by inheritance.
Maybe some kind of proxy can take a place here?

ffried:
@lolgear You mean the whole implementation? That's your decision to make. I'd be fine with a subclass, since most of the work still stays the same. You could create an abstract superclass which does the common work, and two subclasses which do the specific work, but this would require more changes in Client code. With this subclass almost no changes are necessary at all. Actually only if you want to use the new subclass.

lolgear:
@ffried 
No, I mean NSProxy.

```
@interface DDFileLogger ()
- (instancetype)withBuffer; // wrap logger into NSProxy with buffer
- (instancetype)withoutBuffer; // unwrap dependent object from NSProxy.
@end
```

Classes/DDFileLogger.m
lolgear: @ffried 
`- (void)logData` is not only one place for logging data.
`- (void)cleanup` method which is called in dealloc, also should dump data to disk if necessary.

```objective-c
- (void)cleanup {
    [self dumpBufferToDisk];
    [super cleanup];
}
```

will be expanded to

```
- (void)cleanup {
    __auto_type data = (NSData *)[_bufferStream propertyForKey:NSStreamDataWrittenToMemoryStreamKey];
    [super logData:data];
    [self flushBuffer];
    [super cleanup];
}
```

About locking.
Is it necessary for file logger?
every logger, as I know, runs inside serial queue, no?

Consider `lt_log:`

```
    if (_numProcessors > 1) {
        // Execute each logger concurrently, each within its own queue.
        // All blocks are added to same group.
        // After each block has been queued, wait on group.
        //
        // The waiting ensures that a slow logger doesn't end up with a large queue of pending log messages.
        // This would defeat the purpose of the efforts we made earlier to restrict the max queue size.

        for (DDLoggerNode *loggerNode in self._loggers) {
            // skip the loggers that shouldn't write this message based on the log level

            if (!(logMessage->_flag & loggerNode->_level)) {
                continue;
            }
            
            dispatch_group_async(_loggingGroup, loggerNode->_loggerQueue, ^{ @autoreleasepool {
                [loggerNode->_logger logMessage:logMessage];
            } });
        }
        
        dispatch_group_wait(_loggingGroup, DISPATCH_TIME_FOREVER);
    } else {
        // Execute each logger serialy, each within its own queue.
        
        for (DDLoggerNode *loggerNode in self._loggers) {
            // skip the loggers that shouldn't write this message based on the log level

            if (!(logMessage->_flag & loggerNode->_level)) {
                continue;
            }
            
            dispatch_sync(loggerNode->_loggerQueue, ^{ @autoreleasepool {
                [loggerNode->_logger logMessage:logMessage];
            } });
        }
    }

```

There is serial logging ( with group or with sync ) in both branches.
However, yes, If logger can be added to `loggers` array twice or same loggers could write to same file, some problems could occur.

ffried:
@lolgear Again, your decision. I'm fine with either solution. I'd prefer one that needs little changes to existing client code, but if it's otherwise a lot nicer, I'm fine with it, too.

Classes/DDFileLogger.m
ffried: I see. What if we add a method to `DDFileLogger` called `- (void)writeToFile:(NSData *)data` wich does what `- (void)logData:(NSData *)data` currently does? The latter would then just call the former, but the subclass could then simply call `[self writeToFile:data];` instead of "misusing" `super` for this.

About locking: You might be right. I just recalled, that we recently added locking around some properties (e.g. logFileHandle) in DDFileLogger.

lolgear:
@ffried 
Could you compare both implementations?
I think that Proxy is better in terms of both code style and decoupling.
Also this Proxy can be extracted into Extensions and it can wrap any logger.

Classes/DDFileLogger.m
```diff
@@ -1531,3 +1540,231 @@ BOOL doesAppRunInBackground() {
...
+        }
+        const uint8_t *appendedData = calloc(length, sizeof(uint8_t));
+        [data getBytes:(void *)appendedData length:length];
+        if (appendedData != NULL) {
+            [_bufferStream write:appendedData maxLength:length];
+        }
+        if (appendedData != NULL) {
+            free((void *)appendedData);
+        }
+        _bufferSize += _bufferSize + length;
```
ffried: Isn't this increasing `_bufferSize` too much? This should be `_bufferSize += length`, right?

Classes/DDFileLogger.m
```diff
@@ -1531,3 +1540,231 @@ BOOL doesAppRunInBackground() {
...
+
+- (instancetype)initWithInstance:(DDFileLogger *)instance {
+    self.instance = instance;
+    self.maximumBytesCountInBuffer = kDefaultBytesCountInBuffer;
+    return self;
+}
+
+- (void)dealloc {
+    [self dumpBufferToDisk];
+    self.instance = nil;
```
ffried: This isn't necessary, right? All properties will be set to `nil` on dealloc anyways.

ffried:
Ok, looking at the warnings I need to change my decision. We cannot use `NSProxy` with `instancetype` (in `wrapWithBuffer` and `unwrapFromBuffer`), because we don't really return an instance of `DDFileLogger` there. This _might_ lead to problems (especially with Swift) if the compiler makes some assumptions...

Classes/DDFileLogger.m
lolgear: @ffried 
Yes, I mixed both statements here :)

Classes/DDFileLogger.m
lolgear: @ffried 
Sure, it is an explicit assignment to specify order of dealloc.
It could be removed.

lolgear:
@ffried 
Well, we can use explicit type or type conversion in wrapping methods.
If it is only one "corner stone" between proxy and lumberjack, maybe we can move it away?

ffried:
@lolgear: Ok, I've just played around with NSProxy in a mixed-language project. It seems like adding an explicit typecast will do the trick without breaking `is` checks and `as?/!` casts in Swift.

ffried:
@lolgear Can you remove the subclass-version and just leave the proxy version? I'll then have another look at the code, so we can finalize this MR.
Can you also update the Changelog please?

Classes/Extensions/DDFileLogger+Buffering.m
```diff
@@ -0,0 +1,165 @@
...
+//   to endorse or promote products derived from this software without specific
+//   prior written permission of Deusty, LLC.
+
+#import "DDFileLogger+Buffering.h"
+
+@interface DDFileLogger (Internal)
+- (void)logData:(NSData *)data;
+@end
+
+// TODO: Fix it later. ( Protocol adoption? )
```
ffried: Can we simply put this category into its own file combination (.h/.m) and also implement the `-writeToFile:` method there? This would prevent this TODO...

Classes/Extensions/DDFileLogger+Buffering.m
```diff
@@ -0,0 +1,165 @@
...
+#pragma clang diagnostic ignored "-Wincomplete-implementation"
+@implementation DDFileLogger (Internal)
+@end
+#pragma clang diagnostic pop
+
+static NSUInteger kMaximumBytesCountInBuffer = (1 << 10) * (1 << 10); // 1 MB.
+static NSUInteger kDefaultBytesCountInBuffer = (1 << 10);
+
+// MARK: Public Interface
+@interface DDBufferedProxy<T: DDFileLogger */*id<DDLogger>*/>: NSProxy
```
ffried: Nice idea to make it generic. But since it's not a public proxy (nor should it be right now), I don't think that we should add the complexity of making it generic for now.

ffried:
@lolgear Just like I mentioned in my latest review, I'd put the `-logData:` method and the `-writeToFile:` method in an internal category on `DDFileLogger` for now. Buffering also adds the risk of losing log messages (e.g. when an iOS app goes into background and is later killed). So for now I'd just limit the feature to file loggers until we see how they're used and what issues come up. Once it works fine, we can still always expand it to other loggers. The approach we're taking here makes this easy. But we should try to take one step at a time... 😉

Classes/Extensions/DDFileLogger+Buffering.m
lolgear: @ffried 
`-writeToFile:` uses `-currentLogFileHandle` which is defined in `DDFileLogger.m`. So, it is impossible to separate them.
Maybe we could add additional protocol or additional method to existing protocol and marked it `optional`(`-logData` will fit it excellently)?

Classes/Extensions/DDFileLogger+Buffering.m
ffried: If we keep it, we should do it right, at least:
- `T` should be named `FileLogger`
- `/*id<DDLogger>*/` can be removed to make it more readable.
- In the implementation, we should then also use the generic type and not `DDFileLogger`.

ffried: AFAIK they're raised to the highest possible type. In our case `DDFileLogger`. So the `typedef` is not necessary. Since we don't use it with a subclass of `DDFileLogger`, using `__kindof` will not change anything. Maybe even complicate compilation.

Classes/Extensions/DDFileLogger+Buffering.m
```diff
@@ -0,0 +1,155 @@
...
+}
+
+- (void)dealloc {
+    [self dumpBufferToDisk];
+    self.instance = nil;
+}
+
+#pragma mark - Logging
+- (void)logData:(NSData *)data {
+    if ([self isBufferFull]) {
```
c0diq: This check would be better done after the `appendToBuffer` so that it gets flushed immediately when full rather than wait for the next `logData` call.

PR #1003, opened 2018-12-27
DDFileLogger queues and refactoring

hhanesand:
- Use already provided queue to manage access to variables in file logger. The queue is created in the abstract superclass as a serial queue, which fits our needs.
- Prefix internal methods meant to be called on internal queue and add asserts to avoid further threading issues.
- Refactor out some if statements determining which log files to archive.
- Add DISPATCH_VNODE_REVOKE to vnode observation flags.
- Remove usage of `fileExistsAtPath:`, instead just attempt the operation. withIntermediateDirectories: YES doesn't return error if directory exists.

### DIFF
--- a/Classes/DDFileLogger.h
+++ b/Classes/DDFileLogger.h
@@ -302,9 +302,7 @@ extern unsigned long long const kDDDefaultLogFilesDiskQuota;
 /**
  *  The standard implementation for a file logger
  */
-@interface DDFileLogger : DDAbstractLogger <DDLogger> {
-	DDLogFileInfo *_currentLogFileInfo;
-}
+@interface DDFileLogger : DDAbstractLogger <DDLogger>
 
 /**
  *  Default initializer
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -263,15 +263,13 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
     // We could do this check once, during initalization, and not bother again.
     // But this way the code continues to work if the directory gets deleted while the code is running.
 
-    if (![[NSFileManager defaultManager] fileExistsAtPath:_logsDirectory]) {
-        NSError *err = nil;
-
-        if (![[NSFileManager defaultManager] createDirectoryAtPath:_logsDirectory
-                                       withIntermediateDirectories:YES
-                                                        attributes:nil
-                                                             error:&err]) {
-            NSLogError(@"DDFileLogManagerDefault: Error creating logsDirectory: %@", err);
-        }
+    NSError *err = nil;
+    BOOL success = [[NSFileManager defaultManager] createDirectoryAtPath:_logsDirectory
+                                             withIntermediateDirectories:YES
+                                                              attributes:nil
+                                                                   error:&err];
+    if (success == NO) {
+        NSLogError(@"DDFileLogManagerDefault: Error creating logsDirectory: %@", err);
     }
 
     return _logsDirectory;
@@ -389,7 +387,8 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
 }
 
 - (NSArray *)sortedLogFileInfos {
-    return  [[self unsortedLogFileInfos] sortedArrayUsingComparator:^NSComparisonResult(DDLogFileInfo   * _Nonnull obj1, DDLogFileInfo   * _Nonnull obj2) {
+    return  [[self unsortedLogFileInfos] sortedArrayUsingComparator:^NSComparisonResult(DDLogFileInfo *obj1,
+                                                                                        DDLogFileInfo *obj2) {
         NSDate *date1 = [NSDate new];
         NSDate *date2 = [NSDate new];
 
@@ -549,30 +548,18 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 @interface DDFileLogger () {
-    __strong id <DDLogFileManager> _logFileManager;
-    
+    id <DDLogFileManager> _logFileManager;
+
+    DDLogFileInfo *_currentLogFileInfo;
     NSFileHandle *_currentLogFileHandle;
     
     dispatch_source_t _currentLogFileVnode;
+
+    NSTimeInterval _rollingFrequency;
     dispatch_source_t _rollingTimer;
     
     unsigned long long _maximumFileSize;
-    NSTimeInterval _rollingFrequency;
 }
-
-/**
- Serial queue for synchronous access to `currentLogFileHandle`.
- */
-@property (nonatomic, strong) dispatch_queue_t currentLogFileHandleQueue;
-
-/**
- Serial queue for synchronous access to `currentLogFileInfo`.
- */
-@property (nonatomic, strong) dispatch_queue_t currentLogFileInfoQueue;
-
-- (void)rollLogFileNow;
-- (void)maybeRollLogFileDueToAge;
-- (void)maybeRollLogFileDueToSize;
 
 @end
 
@@ -580,7 +567,6 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
 
 - (instancetype)init {
     DDLogFileManagerDefault *defaultLogFileManager = [[DDLogFileManagerDefault alloc] init];
-
     return [self initWithLogFileManager:defaultLogFileManager];
 }
 
@@ -589,13 +575,9 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
         _maximumFileSize = kDDDefaultLogMaxFileSize;
         _rollingFrequency = kDDDefaultLogRollingFrequency;
         _automaticallyAppendNewlineForCustomFormatters = YES;
-        
-        _currentLogFileHandleQueue = dispatch_queue_create("cocoa.lumberjack.fileLogger.currentLogFileHandleQueue", DISPATCH_QUEUE_SERIAL);
-        _currentLogFileInfoQueue = dispatch_queue_create("cocoa.lumberjack.fileLogger.currentLogFileInfoQueue", DISPATCH_QUEUE_SERIAL);
-        
-        logFileManager = aLogFileManager;
 
-        self.logFormatter = [DDLogFileFormatterDefault new];
+        _logFileManager = aLogFileManager;
+        _logFormatter = [DDLogFileFormatterDefault new];
     }
 
     return self;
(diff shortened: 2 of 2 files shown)


Classes/DDFileLogger.h
```diff
@@ -302,9 +302,7 @@ extern unsigned long long const kDDDefaultLogFilesDiskQuota;
 /**
  *  The standard implementation for a file logger
  */
-@interface DDFileLogger : DDAbstractLogger <DDLogger> {
-	DDLogFileInfo *_currentLogFileInfo;
-}
+@interface DDFileLogger : DDAbstractLogger <DDLogger>
```
hhanesand: This absolutely has to be protected by the logger queue, and hence we can't expose it here in the header. It's been public since 2016, so this is a breaking change.

Classes/DDFileLogger.m
```diff
@@ -263,15 +263,13 @@ - (NSString *)logsDirectory {
...
-                                                             error:&err]) {
-            NSLogError(@"DDFileLogManagerDefault: Error creating logsDirectory: %@", err);
-        }
+    NSError *err = nil;
+    BOOL success = [[NSFileManager defaultManager] createDirectoryAtPath:_logsDirectory
+                                             withIntermediateDirectories:YES
+                                                              attributes:nil
+                                                                   error:&err];
+    if (success == NO) {
+        NSLogError(@"DDFileLogManagerDefault: Error creating logsDirectory: %@", err);
```
hhanesand: For this, see the comment above `-[NSFileManager fileExistsAtPath]`

OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Classes/DDFileLogger.m
```diff
@@ -827,23 +812,21 @@ - (void)rollLogFileNow {
...
 
+    _currentLogFileInfo.isArchived = YES;
+    NSString *archivedFilePath = [_currentLogFileInfo.filePath copy];
     _currentLogFileInfo = nil;
 
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        if ([self->_logFileManager respondsToSelector:@selector(didRollAndArchiveLogFile:)]) {
+            [self->_logFileManager didRollAndArchiveLogFile:archivedFilePath];
+        }
+    });
```
hhanesand: adding a dispatch_async here prevents deadlock if _logFileManager logs an error

Classes/DDFileLogger.m
```diff
@@ -884,128 +870,153 @@ - (void)maybeRollLogFileDueToSize {
...
+                    [self->_logFileManager didArchiveLogFile:archivedLogFilePath];
                 }
-            }
-
-            if (self->_currentLogFileInfo == nil) {
-                NSString *currentLogFilePath = [self.logFileManager createNewLogFile];
-
-                self->_currentLogFileInfo = [[DDLogFileInfo alloc] initWithFilePath:currentLogFilePath];
-            }
+            });
```
hhanesand: adding a dispatch_async here prevents deadlock if _logFileManager logs an error

Classes/DDFileLogger.h
```diff
@@ -302,9 +306,7 @@ extern unsigned long long const kDDDefaultLogFilesDiskQuota;
 /**
  *  The standard implementation for a file logger
  */
-@interface DDFileLogger : DDAbstractLogger <DDLogger> {
-	DDLogFileInfo *_currentLogFileInfo;
-}
```
hhanesand: This absolutely has to be protected by the logger queue, and hence we can't expose it here in the header. It's been public since 2016, so this is a breaking change.

Classes/DDFileLogger.m
```diff
@@ -905,128 +933,152 @@ - (void)maybeRollLogFileDueToSize {
...
-
-                __weak __auto_type weakSelf = self;
-                dispatch_source_set_event_handler(self->_currentLogFileVnode, ^{ @autoreleasepool {
-                    NSLogInfo(@"DDFileLogger: Current logfile was moved. Rolling it and creating a new one");
-                    [weakSelf rollLogFileNow];
-                } });
-                
+- (void)lt_monitorCurrentLogFileForExternalChanges {
+    NSAssert(self->_currentLogFileHandle, @"Can not monitor without handle.");
+    NSAssert([self isOnInternalLoggerQueue], @"lt_ methods should be on logger queue.");
```
ffried: Nit-Pick: Logically we should first assert being on the internal queue and then access the currentLogFileHandle.

Classes/DDFileLogger.m
```diff
@@ -1044,18 +1096,24 @@ - (void)logMessage:(DDLogMessage *)logMessage {
     }
 
     if (message) {
-        if ((!isFormatted || _automaticallyAppendNewlineForCustomFormatters) &&
-            (![message hasSuffix:@"\n"])) {
```
ffried: Like above, I think we're fine with expecting `\n` to be the newline character here.

Classes/DDFileLogger.m
```diff
@@ -1077,27 +1135,26 @@ - (void)willLogMessage {
...
 }
 
 - (void)willRemoveLogger {
-    // If you override me be sure to invoke [super willRemoveLogger];
+    [self lt_rollLogFileNow];
+}
 
-    [self rollLogFileNow];
+- (void)flush {
+    NSAssert([self isOnInternalLoggerQueue], @"lt_ methods should be on logger queue.");
```
ffried: Assertion method is incorrect, it's not an `lt_` method...

Classes/DDFileLogger.m
```diff
@@ -391,7 +402,8 @@ - (NSArray *)sortedLogFileNames {
 }
 
 - (NSArray *)sortedLogFileInfos {
-    return  [[self unsortedLogFileInfos] sortedArrayUsingComparator:^NSComparisonResult(DDLogFileInfo   * _Nonnull obj1, DDLogFileInfo   * _Nonnull obj2) {
+    return  [[self unsortedLogFileInfos] sortedArrayUsingComparator:^NSComparisonResult(DDLogFileInfo *obj1,
```
ffried: Nit-Pick: Two spaces after `return`.

Classes/DDFileLogger.h
```diff
@@ -310,19 +316,23 @@ extern unsigned long long const kDDDefaultLogFilesDiskQuota;
...
+@interface DDFileLogger : DDAbstractLogger <DDLogger>
 
 /**
- *  Default initializer
+ *  Default initializer.
  */
 - (instancetype)init;
 
+- (instancetype)initWithLogFileManager:(id <DDLogFileManager> __nullable)logFileManager
+        __attribute__((deprecated("Use -initWithLogFileManager:completionQueue:")));
```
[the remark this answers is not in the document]
hhanesand: I was worried about visibility here. The warning would point out the change and introduce the new initializer. What do you think? I'm open to both.

Classes/DDFileLogger.m
hhanesand: when withIntermediateDirectories:YES is specified, the FileManager does not return any errors if the directory already exists.

Doc for return value on that method :
> true if the directory was created, true if createIntermediates is set and the directory already exists, or false if an error occurred.

https://developer.apple.com/documentation/foundation/filemanager/1407884-createdirectory

Classes/DDFileLogger.m
```diff
@@ -429,32 +442,46 @@ - (NSString *)newLogFileName {
...
+    if (fileHeaderStr.length == 0) {
         return nil;
     }
     
     // Ensure that we have a newline at the end of the string
-    fileHeaderStr = [NSString stringWithFormat:@"%@\n", fileHeaderStr];
-    
-    NSData *fileHeaderData = [fileHeaderStr dataUsingEncoding:NSUTF8StringEncoding];
-    return fileHeaderData;
+    unichar lastChar = [fileHeaderStr characterAtIndex:fileHeaderStr.length - 1];
```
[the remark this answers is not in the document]
hhanesand: yeah, you're right. The logFileHeader shouldn't contain any other newlines

Classes/DDFileLogger.m
hhanesand: This method is currently only called from inside `DDLog lt_flush`, which dispatches onto the logger's queue before calling the method. I was considering renaming it, but I believe that would be another breaking change since the method is defined in the public `DDLogger` protocol. I therefore opted to insert the check and not rename the method. I'll insert a comment since this is rather non-intuitive.

Classes/DDFileLogger.h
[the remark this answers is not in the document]
ffried: I know many developers (me included) who work with the "Treat warnings as errors" compiler setting enabled. Those will hate us for this. 😉 
Also, those that really need this new initializer will find it. Those who don't, anyways don't care and are happy if there's neither a warning nor an error.
But what you can do, to improve visibility is to add a header doc to this initializer. People working with Xcode's quick help will see it, as well as those looking at the headers.

Classes/DDFileLogger.m
hhanesand: good catch

Classes/DDFileLogger.m
ffried: A comment would most certainly help, yea.
But why deprecate `flush` in your alternative? Why not just add `lt_flush` and forward `flush` -> `lt_flush`? People calling `flush` from outside the logger probably *want* it to be safe (as in dispatched to the internal queue), right?

Classes/DDFileLogger.m
ffried: Ah, you're right. Thanks for the heads up... 🙂

Classes/DDFileLogger+Internal.h
```diff
@@ -18,8 +18,9 @@
 NS_ASSUME_NONNULL_BEGIN
 
 @interface DDFileLogger (Internal)
-- (void)logData:(NSData *)data;
-- (void)writeToFile:(NSData *)data;
+
+- (void)lt_logData:(NSData *)data;
+
 @end
```
hhanesand: Saw no reason to have logData calling writeToFile, when writeToFile is not used anywhere else in the project.

Classes/DDFileLogger.h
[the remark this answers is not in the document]
hhanesand: Went ahead and added the header docs.

Classes/DDFileLogger.m
hhanesand: Yea, let's make flush save for outside use 👍

Classes/Extensions/DDFileLogger+Buffering.m
```diff
@@ -64,15 +70,14 @@ - (void)appendToBuffer:(NSData *)data {
...
-        [data getBytes:(void *)appendedData length:length];
+        const uint8_t *appendedData = calloc(length, sizeof(uint8_t));
         if (appendedData != NULL) {
+            [data getBytes:(void *)appendedData length:length];
             [_bufferStream write:appendedData maxLength:length];
-        }
-        if (appendedData != NULL) {
+            _bufferSize += length;
+
             free((void *)appendedData);
```
hhanesand: Made a slight semantic change here.

hhanesand:
@ffried I've addressed comments from your review and merged master. Looked over the buffering code and it should play fine with the queue changes. I'll test this branch in one of my projects today.

ffried:
@hhanesand thanks! I’ll take another look at the code tomorrow. Once you give the green light after your tests, we can merge this. 👍

PR #1014, opened 2019-01-15
Spelling.

hhanesand:
Also removes some unneeded imports and changes a 0 -> nil.


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Copyright is not valid. See our default copyright in all of our files (Classed and Demos use different format).</td>
    </tr>
  </tbody>
</table>

Invalid copyright: Benchmarking/PerformanceTesting.m

</p>

.idea/dictionaries/hhanesand.xml
```diff
@@ -0,0 +1,78 @@
+<component name="ProjectDictionaryState">
```
lolgear: Is it local file from JetBrains that should be excluded?

hhanesand: It's a list of accepted spellings. It will be picked up by anyone using AppCode. (not just for me even though it has my name in it) I've double checked this by renaming the name inside the dictionary, and it still loads the file.

lolgear: I assume that .gitignore should contain all .idea related files as all user-related Xcode files as any IDE-related user files.

hhanesand:
@ffried I did a project wide replace, so all occurrences should be fixed. Seems like the templates have the correct spelling.

ffried:
@hhanesand I just double checked. The templates are indeed correct. So I only messed up copy-pasting it in all the source files. 🙃

.idea/dictionaries/hhanesand.xml
hhanesand: Was trying to articulate that I think this file is of use to people other than me, and therefore we shouln't have it in the `.gitignore`. Do you disagree?

ffried: Maybe we can just rename the dictionary to `cocoalumberjack.xml` then? I agree that _user-related_ IDE files should be excluded. But this is not _user-related_ but rather _project-related_ (like the contents of `xcshareddata` in a xcodeproj). So from that point of view this file is fine.

However, we do depend on certain Xcode features (like the `IDETemplateMacros.plist` file) of which I don't know if AppCode has them as well. So the question here is more: Should we officially support AppCode, or should we just passively do so.

hhanesand: I vote for passively supporting, too much work otherwise. There is a reference in the repository already for AppCode: `./Documentation/AppCode-support.md`.

PR #1015, opened 2019-01-16
Travis Build Matricies

hhanesand:


hhanesand:
Maybe the build isn't getting triggered because the .travis.yml file is invalid. Not sure.

ffried:
@hhanesand you can validate the YAML as described [here](https://support.travis-ci.com/hc/en-us/articles/115002904174-Validating-travis-yml-files).

lolgear:
@hhanesand 
We can go further and add script inside travis that generates build commands.
```yaml
...
                - xcodebuild build -workspace Demos/Demos.xcworkspace -scheme 'NonArcTest' | xcpretty -c
                - xcodebuild build -workspace Demos/Demos.xcworkspace -scheme 'PerUserLogLevels' | xcpretty -c
```

Dreadful picture :(

hhanesand:
Alright, so those are passing now. We've gone from ~14 minutes -> ~10 minutes, so not too much of a speedup, but we are testing more ios and mac versions now.

I've also configured the demo apps to only be built while pushing on the master branch when we release, since they take a while.

https://travis-ci.org/CocoaLumberjack/CocoaLumberjack/builds/480703606

Classes/DDASLLogCapture.m
```diff
@@ -35,7 +35,7 @@
 @implementation DDASLLogCapture
 
 + (void)start {
-    // Ignore subsequent calls
+    // Ignore subsequent callss
```
ffried: We should revert this 😉

.travis.yml
```diff
@@ -1,89 +1,64 @@
...
+    include:
+        - stage: compile_core
+          script: bash ./Scripts/compile_core.sh
 
-after_success:
-    - bash <(curl -s https://codecov.io/bash)
+        - stage: compile_demo_apps
+          if: env(TRAVIS_PULL_REQUEST) = false
+          script: bash ./Scripts/compile_demo_apps.sh
+
```
ffried: NitPick: Missing newline at the end of file

Gemfile
```diff
@@ -2,3 +2,5 @@ source "https://rubygems.org"
 
 gem 'danger'
 gem 'danger-swiftlint'
+gem 'cocoapods'
+gem 'xcpretty'
```
ffried: NitPick: Missing newline at the end of file

Scripts/compile_core.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+set -e # exit on first error, and print commands with expansions
```
ffried: I usually use `set -euxo pipefail` in my scripts:
```
# -e: Fail if command fails
# -u: Fail on undefined vars
# -x: Print commands
# -o pipefail: check each command's exit code in a pipe
```

ffried:
@hhanesand Oh, and there seem to be conflicts with `master`.

Scripts/compile_core.sh
```diff
@@ -0,0 +1,59 @@
...
+SDK="appletvsimulator$OS"
+NAME="Apple TV"
+PLATFORM="tvOS Simulator"
+build_project "" "Integration/Integration.xcodeproj" "tvOSSwiftIntegration"
+
+OS="$DEFAULT_MAC_OS"
+SDK="macOS$OS"
+NAME=""
+PLATFORM="macOS"
+build_project "" "Integration/Integration.xcodeproj" "macOSSwiftIntegration"
```
ffried: NitPick: Missing newline at the end of file

hhanesand:
That lib you found looks nice, I'll toss it in tomorrow.

lolgear:
@hhanesand @ffried @bpoplauschi 
I am a bit confused about `.sh` as storage of actions.
It is too difficult to update if somebody would like to do it. ( ahhhh, If-statements are the messssshhh )
Since it is AppleOS library, I would like to suggest most common tool for it - ruby + 'shellwords' gem. All macOS modern tools ( fastlane/brew/cocoapods ) are written in Ruby.

What do you think?

Scripts/compile_demo_apps.sh
```diff
@@ -0,0 +1,27 @@
...
+build() {
+    xcodebuild build                       \
+        -workspace Demos/Demos.xcworkspace \
+        -scheme "$1"                       \
+        -sdk iphonesimulator               \
+    | bundle exec xcpretty -c
+}
+
+# tail -n +3 removes the first 3 lines
+SCHEMES=$(xcodebuild -workspace Demos/Demos.xcworkspace/ -list | tail -n +3)
```
lolgear: Is it correct? Search for `Schemes:` as I suggest will be more safe. also, as I suppose, `head` can skip first lines.

Scripts/compile_demo_apps.sh
```diff
@@ -0,0 +1,27 @@
...
+    # remove leading whitespace characters
+    var="${var#"${var%%[![:space:]]*}"}"
+    # remove trailing whitespace characters
+    var="${var%"${var##*[![:space:]]}"}"   
+    echo -n "$var"
+}
+
+build() {
+    xcodebuild build                       \
+        -workspace Demos/Demos.xcworkspace \
```
lolgear: Use one style. I suggest to wrap every path in "" since path can contain whitespaces. You do so in Tests script.

Scripts/compile_core.sh
```diff
@@ -0,0 +1,59 @@
...
+build_workspace "clean" "Framework/Lumberjack.xcworkspace" "CocoaLumberjack"
+build_workspace "clean" "Framework/Lumberjack.xcworkspace" "CocoaLumberjackSwift"
+
+build_project "" "Integration/Integration.xcodeproj" "iOSStaticLibraryIntegration"
+build_project "" "Integration/Integration.xcodeproj" "iOSFrameworkIntegration"
+build_project "" "Integration/Integration.xcodeproj" "iOSSwiftIntegration"
+
+OS="$$DEFAULT_WATCH_OS"
+SDK="watchSimulator$OS"
+build_project "" "Integration/Integration.xcodeproj" "watchOSSwiftIntegration"
```
lolgear: `echo "Empty NAME only allowed on macOS."`
But here you don't set NAME for `watchOS` simulator.

Scripts/compile_demo_apps.sh
lolgear: Also, how do you map scheme on its sdk/target? Not all schemes are for iphonesimulator. There are some schemes for macOS.

ffried:
@lolgear 
> Python.. ah, no, writing everywhere self as first parameter for function and different number of tabs for all statements - no, of course not for scripts as part of mature framework.

Many parts of Swift's build system are written in Python. Or take a look at TensorFlow. So much for "scripts as part of a mature framework". 😉 

But again, I see Python and Ruby on the same level (you might as well at Go and/or Swift to that list). And they all have their advantages and disadvantages.

Talking about DSLs: We use the DSL Travis provides. We hit its limits and we moved what hit the limits into the next closest thing: scripts. Previously we had everything in the .travis.yml file and it was kinda "shell-ish" there as well.

But again: this is all my personal opinion. If there's a team decision to use Ruby (or any other higher level language) here, I'm ok with that too.

ffried:
@lolgear 
> However, can we just put all data for script in json or in yaml files? It will increase testing ability and "adding new task entry" will be easy.

That, however, would introduce a new DSL just for our CI builds. And this would certainly introduce more complexity into the scripts (parsing and validating the contents of the YAML/JSON file). We could maybe adjust the scripts a little so that their "main"-code is in the .travis.yml and we extract the repetitive work into the script, passing in the necessary parameters (it's what I do in my personal OSS projects that are tested with Travis).

lolgear:
@ffried 
Now we have two parts of "DSL" with storing some parts of it in Travis and in script.
If Travis really "parallelize" its matrix, it is ok, but, passing parameters as "env" variables is the "dreadfulness".

If I would like to build this script locally, **which options** should I follow?
( Remember, Travis and script have different parts of data ).

ffried:
@lolgear 
> If I would like to build this script locally, which options should I follow?
( Remember, Travis and script have different parts of data ).

How do you run a .travis.yml locally? Same problem. These scripts aren't made to be run locally. If a command fails, I'd do the same as I do today: Look in travis what was executed and run that command locally.

And ENV-Variables are how Travis works. We could maybe just pass them down to the scripts in the .travis.yml directly, yea.

lolgear:
@ffried 
For that we need a script that takes arguments from travis and passes them into another script.

Scripts/compile_demo_apps.sh
hhanesand: xcodebuild handles it I believe. The commands we used earlier to build the demo projects didn't have a `-destination` specifier.

hhanesand: Regarding `tail`, I think it's alright enough. It is correct the way it's implemented now. It's either that or parsing the json returned by `xcodebuild -workspace Demos/Demos.xcworkspace/ -list -json`.

Scripts/compile_core.sh
hhanesand: The watch os platform runs on an iOS simulator, and `$NAME` is already set by the build matrix to "iPhone 8".

lolgear: @hhanesand 
And that is why I don't like environment variables :(

Scripts/compile_demo_apps.sh
lolgear: @hhanesand json is better. I caught strange output several times:

```
$ xcodebuild -workspace Demos.xcworkspace/ -list
2019-01-16 10:25:20.573 xcodebuild[71745:2759467]  DTDeviceKit: deviceType from 667f7238dd76a8f087e96e9cc171729a217cc561 was NULL
2019-01-16 10:25:20.688 xcodebuild[71745:2759464]  DTDeviceKit: deviceType from 667f7238dd76a8f087e96e9cc171729a217cc561 was NULL
Information about workspace "Demos":
    Schemes:
        BenchmarkIPhone
```

hhanesand:
Man, I really wish the -destination specifier was a little easier to handle. Something like a "I don't care what device you pick as long as it works and has this OS version" flag would be super nice. Having to specify the `name` as `iPhone XS` forces anyone who wants to use the xcodebuild to have a iPhone XS WITH a watch attached...

hhanesand:
failing, see https://github.com/Microsoft/azure-pipelines-tasks/issues/8840

ffried:
@hhanesand Huh, some failures seem to be because of a broken command:
`xcodebuild test -project Tests/Tests.xcodeproj -scheme 'OS X Tests' -sdk macOS -destination 'platform=macOS,OS=10.14,name=iPhone XS'` can't work, can it?

Scripts/test.sh
```diff
@@ -0,0 +1,24 @@
...
+source ./Scripts/setup_default_env.sh
+
+test() {
+	echo "Testing $1"
+
+	xcodebuild test                                                         \
+		-project "Tests/Tests.xcodeproj"                                    \
+	 	-scheme "$1"                                                        \
+	 	-sdk "$SDK"                                                         \
+	 	-destination platform="$PLATFORM",OS="$OS",name="$NAME"             \
```
sushichop: OS and name are not needed(must be removed) for `OS X Tests`.

sushichop:
Good works😃
Then, as I commented above, I think OS and name should be removed from macOS destination. 
For example, expected settings are here.

```
platform=macOS
platform=iOS Simulator,OS=9.3,name=iPhone 5s
```

sushichop:
Good! Travis CI failure has been fixed😃
Last one. File conflicts need to be fixed.

Issue #665, opened 2015-12-01
Synchronous logging in Swift?


mark-anders:
Is it possible to enable synchronous logging globally when using Swift? All links that I've seen reference doing it with macros.

ffried:
This is actually an interesting idea. It could be solved similar to the log level in #952.

StevenMagdy:
Is there any way to currently achieve this?

ffried:
@StevenMagdy Not _currently_, but with #1019 will be 😉

bpoplauschi:
Fixed by #1019

Issue #1009, opened 2019-01-11
Fix severity level of stale bot


lolgear:
Hm, It seems that @stalebot eats too much.
- Feature Request
- Discussion (no, of course. It is a forum post and not an issue. It can be closed only by )
- Awaiting Response ( ok, here we have fifty-fifty probability that issue was fixed and it is super rare, it can be closed ).

## Feature Request
These issues show a vector of future of this project.
Only a person can close this issue, not a bot.

## Discussion
These issues are not related to bug reporting. It is a general forum post with discussion and only for people.
Again, only a person can close this issue, not a bot.
Closing this issue by a bot is pinning "Shut your mouth" sign to community mouth.

## Awaiting Response
50/50.
This issue could be fixed and person is unreachable OR it is super rare and person is still unreachable.

What do you think?

ffried:
**Feature request**
Yep, those might be worth excluding.

**Discussion**
I'm not a big fan of having those in issues. But on GitHub itself it seems like the only easy way to have those. Unless we have an open team for those.
We could exclude them from stale-bot. But who decides then if the discussion ended nowhere and should be closed? Who will do this job regularly? Having stale-bot slap the Stale label on them makes us aware of stale discussions. We can then still "un-stale" them by adding a simple comment, that might revive the discussion. If not, maybe it's not the wrong thing to close it. Also, why do we fear closed issues? No one says they cannot be re-opened...

**Awaiting Response**
Same as with discussions. If the OP does not respond within a given timeframe, the issue does not seem to have enough importance. For issues that seem important to use, we can still double the available timeframe by adding a simple comment.

lolgear:
Also, all pinned issues should be invincible.

ffried:
@lolgear Pinned issues *are* already left alone by stale-bot...

As for the other options:
- We currently don't use projects, so I'm not sure this is necessary - but it therefore wouldn't hurt, either.
- Milestones: This definitively sounds good.
- Assignees: AFAIK, anyone can assign issues on GitHub. So IMHO this is not really a good idea.

lolgear:
@ffried 
So, it is another PR for GitHub team. :)
They still don't have starred repositories history, heh :/ I can unstar issue and forget about it :/

PR #1019, opened 2019-01-21
Add global variable to control async logging in Swift

ffried:

This merge request fixes / refers to the following issues: #665 

### Pull Request Description

This adds a global variable in Swift (`asyncLoggingEnabled`) which is then used in all log-functions (except `DDLogError`) to define whether logs are written asynchronously or not.

### DIFF
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -13,8 +13,6 @@
 //   to endorse or promote products derived from this software without specific
 //   prior written permission of Deusty, LLC.
 
-import Foundation
-
 extension DDLogFlag {
     public static func from(_ logLevel: DDLogLevel) -> DDLogFlag {
         return DDLogFlag(rawValue: logLevel.rawValue)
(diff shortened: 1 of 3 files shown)


Classes/CocoaLumberjack.swift
```diff
@@ -71,38 +69,100 @@ public func resetDefaultDebugLevel() {
     resetDynamicLogLevel()
 }
 
+/// If `true`, all logs (except errors) are logged asynchronously by default.
+public var asyncLoggingEnabled = true
```
lolgear: Too ObjC-ish. Not enough nested classes/generics for Swiftish way :)

However, are you sure that this variable should be so alone?
I suppose that access should be defined by the structure or by the class level.
`AsyncLoggingSettings.Enabled` feels better. And word `Settings` points to something that I am able to change.
However, it doesn't sync access on read/write. Is it ok?

Also, here we can define more complex Swift-ish tricks and whistles - we can define a class with static functions for logging.
And later, we can define a global wrappers that call these functions from this class.

```swift
class DDLogSwiftish {
   // Define settings here.
   class Settings {
       static public var asyncLoggingEnabled = true
    }
   // And functions
   @inlineable public static verbose(params: [Int]) {}
   // others.
}

@inlinable
public func DDLogVerbose(params: [Int]) {
    DDLogSwiftish.verbose(params: params)
}
```

ffried: Simple question: why introduce a simple new feature so complicated?

`dynamicLogLevel` (formerly known as `defaultDebugLevel`) has been defined like this forever. If you want to know where it belongs to, simply assign it as `CocoaLumberjackSwift.asyncLoggingEnabled = true`. Swift does allow that.

What you describe is a complete refactoring of the Swift interface of CocoaLumberjack. There are a few nice ideas for that, but it's not the goal of this PR. And if we did so, I'd get rid of the global `func`s and DD-prefixes completely and use something like `Log.error(...)`, `Log.verbose(...)`. But this would also be out of scope for 3.5.0 which should be released ASAP.

lolgear: @ffried 
Simple answer - simple decisions today _will_ affect problems tomorrow.
You add one variable that _is_ neither have sync access nor have proper scope ( global scope ). Is it really _good_ practice? I can't remember Cocoas framework which uses it. Why mature opensource use it?

Again, when you define this function, person will `import CocoaLumberjack` as always and this person expects that settings ( various settings, even one setting ) are consolidated in one `facade` ( ah, patterns ) that has name `DDLogSettings`.

ffried: Don't get me wrong. I agree that currently the Swift interface does look a lot like ObjC. I just don't see why I should refactor the whole interface just to implement one small feature.
It's not good Swift practice, no. But it's not always clever to throw everything away just to do something small. If you feel better, I'll open another PR with refactoring of CocoaLumberjackSwift. But I won't stuff it into this one.

But since will be a breaking change (as opposed to this change), it'll have to target a 4.0.0 release.

lolgear: @ffried 
Well, I suggest to update release version of this feature to 4.0.0.

OR

I suggest to rename this variable to `DDLogAsyncLoggingEnabled` (`DDLogSettingsAsyncLoggingEnabled`) to emphasize further refactoring.

ffried: The feature is simple enough and with our release plans, 4.0.0 is IMHO too far away for this feature to wait.

Renaming is IMHO also not the best idea, since it would also mean that `dynamicLogLevel` would need to be renamed and `currentFileName(_:)` probably too. I implemented it so that it matches the current state of the Swift interface. After the refactoring, Swift will tell anyone using the "old" interface, because I'll add the correct deprecation attributes.

lolgear: @ffried 
Hm...
``` dynamicLogLevel ``` for each `import CocoaLumberjack`
Hm...

Ok, if it necessary, so, do it.

File: Classes/CocoaLumberjack.swift [hunk 2]
```diff
@@ -71,38 +69,100 @@ public func resetDefaultDebugLevel() {
...
 @inlinable
-public func DDLogError(_ message: @autoclosure () -> String, level: DDLogLevel = DDDefaultLogLevel, context: Int = 0, file: StaticString = #file, function: StaticString = #function, line: UInt = #line, tag: Any? = nil, asynchronous async: Bool = false, ddlog: DDLog = DDLog.sharedInstance) {
+public func DDLogError(_ message: @autoclosure () -> String,
+                       level: DDLogLevel = DDDefaultLogLevel,
+                       context: Int = 0,
+                       file: StaticString = #file,
+                       function: StaticString = #function,
+                       line: UInt = #line,
+                       tag: Any? = nil,
+                       asynchronous async: Bool = false,
```
lolgear: Should it also have default `asyncLoggingEnabled`?

ffried: Nope. In ObjC, error logging is also always synchronous.

sushichop:
LGTM👍

ffried:
@bpoplauschi The changes look like more than they actually are. I've just split the parameters to the log functions into separate lines making it easier to read the code. The basic change is simply a new global var, that is used as default for each log-function's `asynchronous` parameter.

Since it's a small change with possibly big impact, I'd find it nice if it would make it into 3.5.0 (especially considering the time a release takes to get released 😜). But if it's too much, 3.6.0 is fine with me, too.
It could just happen, that by then I finished the bigger refactoring of the Swift interface (as mentioned [here](https://github.com/CocoaLumberjack/CocoaLumberjack/pull/1019#discussion_r249352646)). 🙂

bpoplauschi:
Got it @ffried. The I will review it before releasing.

bpoplauschi:
I initially thought the impact was bigger.

bpoplauschi:
@CocoaLumberjack/collaborators I want to release 3.5.0 tomorrow. Please let me know if there are any things missing or other considerations. If not, let's finally roll out this bad boy

lolgear:
@bpoplauschi 
I want to add several lines to stale bot as a part of release.
It is not necessary, but, well, it causes several missing issues. ( That are too old, of course, but I insist on less stale bot hunger).
Here I show `exempt*` usage from documentation.
#1009

ffried:
@lolgear @bpoplauschi IMHO, this is not relevant for the release. StaleBot is a repository-configuration. Users of the framework don't need it. Like I just commented in #1009, there are some good ideas there. But they can be merged independently of a release.
Or am I missing any specific reason to put it into a release, @lolgear?

bpoplauschi:
Agree with @ffried

lolgear:
@ffried 
@bpoplauschi 
I can't resist, release :3
But
It costs nothing ( with exempt ) and will prevent further killing of necessary issues ( like sync/async ) logging. And yes, we can assume that this release contains everything for correct work of whole repository. It is like a fresh breath of project. 🧘‍♂️

But, release it. :)
P.S: A bit overwhelmed by interview today :)

PR #1031, opened 2019-01-31
Update Copyrights to 2019 #trivial

### LINKED ISSUE #1030: License file wants back to the future
LICENSE is too old in this year.

However, it is nice to have tool to actualize meta-information like dates, no?

ffried:
Fixes #1030

### DIFF
--- a/Classes/CLI/CLIColor.h
+++ b/Classes/CLI/CLIColor.h
@@ -1,6 +1,6 @@
 // Software License Agreement (BSD License)
 //
-// Copyright (c) 2010-2018, Deusty, LLC
+// Copyright (c) 2010-2019, Deusty, LLC
 // All rights reserved.
 //
 // Redistribution and use of this software in source and binary forms,
--- a/Classes/CLI/CLIColor.m
+++ b/Classes/CLI/CLIColor.m
@@ -1,6 +1,6 @@
 // Software License Agreement (BSD License)
 //
-// Copyright (c) 2010-2018, Deusty, LLC
+// Copyright (c) 2010-2019, Deusty, LLC
 // All rights reserved.
 //
 // Redistribution and use of this software in source and binary forms,
--- a/Classes/CocoaLumberjack.h
+++ b/Classes/CocoaLumberjack.h
@@ -1,6 +1,6 @@
 // Software License Agreement (BSD License)
 //
-// Copyright (c) 2010-2018, Deusty, LLC
+// Copyright (c) 2010-2019, Deusty, LLC
 // All rights reserved.
 //
 // Redistribution and use of this software in source and binary forms,
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -1,6 +1,6 @@
 // Software License Agreement (BSD License)
 //
-// Copyright (c) 2010-2018, Deusty, LLC
+// Copyright (c) 2010-2019, Deusty, LLC
 // All rights reserved.
 //
 // Redistribution and use of this software in source and binary forms,
--- a/Classes/DDASLLogCapture.h
+++ b/Classes/DDASLLogCapture.h
@@ -1,6 +1,6 @@
 // Software License Agreement (BSD License)
 //
-// Copyright (c) 2010-2018, Deusty, LLC
+// Copyright (c) 2010-2019, Deusty, LLC
 // All rights reserved.
 //
 // Redistribution and use of this software in source and binary forms,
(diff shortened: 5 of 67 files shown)


lolgear:
@ffried 
Hm, I am not familiar with date management and licenses management.
But I use file as VERSION with current version of Pod. Then, I just slurp out it in pod spec.

It works for pod spec files since they are just ruby code.

```
class VersionFile
  class << self
    def version
      File.new(self.filepath).gets.rstrip
    end
    def filepath
      './VERSION'
    end
  end
end

Pod::Spec.new do |s|
  s.version      = "#{VersionFile.version}"
```

Some years ago I heard about custom headers for files that Xcode has created. But they are unmanaged after creation.
Setting __YEAR__ constant works IF we could managed it by specifying in `LICENSE_DATE` file, for example.
Otherwise it is just a date of current year and after million days it will show 1000_000/365 = 2019 + 2739 but at GitHub it will show ( last update 2700 years ago ).

Manual updating is a solution, yes...

OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">The Carthage project was modified but CocoaPods podspec wasn’t. Did you forget to update the podspec?</td>
    </tr>
  </tbody>
</table>

</p>

ffried:
@lolgear I'm not sure what you're referring to here. Updating the copyright headers once a year is a no-brainer. Actually, the following command does it for me:
```bash
find -E . -regex ".*\.([hm]|swift|pch)" -exec sed -i"" -e "s|// Copyright (c) 2010-${OLD_YEAR}, Deusty, LLC|// Copyright (c) 2010-${NEW_YEAR}, Deusty, LLC|g" "{}" \;
```

Since I don't want to update the IDE Templates each year, I've used the `___YEAR___` constant there so that new files always get the current year.

With all this, the only two files that need to bed updated manually are `Dangerfile` and `LICENSE` which is fine I think.

lolgear:
@ffried 
Could you save this one-liner into scripts?

ffried:
@lolgear I added a script for updating *all* copyrights (including `Dangerfile` and `LICENSE`).

lolgear:
@ffried 
Nice, but one trivial improvement.

Could you add argument -h ( --help ) to show script description and parameters descriptions with an example at the end.

ffried:
@lolgear done

lolgear:
@ffried 
And last discussable improvement.

Maybe it is better to keep documentation as POD ( Plain old documentation )?
I do the same with perl scripts.

Here is [documentation example for bash](https://www.oreilly.com/library/view/bash-cookbook/0596526784/ch05s02.html).

To check it we can use `perldoc FILENAME` or in script itself `perldoc $0`

ffried:
@lolgear For a script that has less than 70 lines and is called once a year? Don't you think that's a little overkill? What benefit would this documentation give over what `--help` outputs now?
This script doesn't even have any logical branches in its core logic...

ffried:
@lolgear I've personally never used `perldoc` before. But that's just me.

`find` and `sed` are no longer used in a one-liner. `sed` is now extracted into a function that is called by `find`.
But either way, I'd document how to *use* the script and not what tools the script uses. So basically what `--help` outputs. The rest won't have to change anyways until either `find`, `date` or `sed` has breaking changes (which won't happen very soon I guess).

My point is, that writing 100 lines of documentation for 70 lines of code that is used 1 time a year seems like a total overkill to me. Especially considering the fact that said documentation very likely won't be read every year - making it only useful every leap year.

Furthermore, anyone who needs to adjust the script and only knows bash but not perldoc (like me), will have a hard time understanding what actually happens in the script. Also because it's now suddenly a 170 lines file instead of just a 70 lines file.

lolgear:
@ffried 

Pros:
1. Classes/* files are well documented and can be easily integrated into Xcode documentation.
2. This is a internal tool - program for developers from developers.
3. Broken window starts here.

Cons:
1. Too much effort for 70 lines.
2. This is a internal tool - program for developers from developers.
3. Unknown tool `perldoc`.

But I have seen bash magic in another PR where environmental variables are just play their tricks and incorrect builds are running.

However, I suggest it as discussible.
Can be closed I think?

bpoplauschi:
I've just simplified this discussion and merged the PR :) Guys, let's keep it simple

lolgear:
@bpoplauschi 
Bash is not the simplest thing we have ever invented as humanity :)

By the way, could you close connected Issue after merge?

bpoplauschi:
Github automatically closes the connected issues if they are mentioned (in commit message or PR description).

PR #1039, opened 2019-02-05
core: loggers os logger variations have been added.

lolgear:

### Pull Request Description

- OS Logger subsystem and category support has been added.

### Issues
* [x] Check for nil both subsystem and category.
* [ ] non-inline method to get logger.
* [x] setters getters instead of ivars.

### DIFF
--- a/Classes/DDOSLogger.h
+++ b/Classes/DDOSLogger.h
@@ -31,12 +31,20 @@ API_AVAILABLE(macos(10.12), ios(10.0), watchos(3.0), tvos(10.0))
 /**
  *  Singleton method
  *
- *  @return the shared instance
+ *  @return the shared instance with OS_LOG_DEFAULT.
  */
 @property (class, readonly, strong) DDOSLogger *sharedInstance;
 
-@end
-
-@interface DDOSLogger (Variations)
-- (instancetype)initWithSubsystem:(NSString *)subsystem category:(NSString *)category;
+/**
+ Designed initializer
+ 
+ @param subsystem Desired subsystem in log. Consider "org.example"
+ @param category Desired category in log. Consider "Point of interests."
+ @return New instance of DDOSLogger.
+ 
+ @discussion This method accepts parameters of type (String, String)?
+ If both parameters are nil, this method will return logger wrapper for `OS_LOG_DEFAULT`.
+ If both parameters are not nil, it will return logger wrapper for `os_log_create(subsystem, category)`
+ */
+- (instancetype)initWithSubsystem:(NSString *)subsystem category:(NSString *)category NS_DESIGNATED_INITIALIZER;
 @end
--- a/Classes/DDOSLogger.h
+++ b/Classes/DDOSLogger.h
@@ -36,3 +36,8 @@ API_AVAILABLE(macos(10.12), ios(10.0), watchos(3.0), tvos(10.0))
 @property (class, readonly, strong) DDOSLogger *sharedInstance;
 
 @end
+
+@interface DDOSLogger (Variations)
+- (instancetype)withSubsystem:(NSString *)string;
+- (instancetype)withCategory:(NSString *)string;
+@end
--- a/Classes/DDOSLogger.h
+++ b/Classes/DDOSLogger.h
@@ -38,7 +38,5 @@ API_AVAILABLE(macos(10.12), ios(10.0), watchos(3.0), tvos(10.0))
 @end
 
 @interface DDOSLogger (Variations)
-- (instancetype)withSubsystem:(NSString *)string;
-- (instancetype)withCategory:(NSString *)string;
 - (instancetype)initWithSubsystem:(NSString *)subsystem category:(NSString *)category;
 @end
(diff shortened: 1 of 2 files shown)


lolgear:
Could anybody post a review?

lolgear:
@ffried 
According to Apple documentation and [wwdc](https://developer.apple.com/videos/play/wwdc2016/721/), `os_log_create` returns thread-safe singleton object.

```objective-c
os_log_t log = os_log_create("com.your_company.subsystem", "network"); 
// Create thread-safe singleton object that controls behavior of log messages
```

Do we still need to cache it?

Initialization with two parameters - nice idea. But I am confused about their nullability. I suppose that both of them can be nil, but Swift version has `OSLog.init(subsystem: String, category: String)`.

ffried:
@lolgear Didn't know about the singleton. The reason I'd still cache it is that even a singleton needs to be looked up again, right? It'll return the same object for the same subsystem/category combination, yes, but it still needs to look it up.
But this is also not a hard requirement for me. I'm also fine with creating it every time.

As for the nullability:
We need to make both nullable, but the initializer should have an assertion that either both are nil or both are set. In Swift I'd use an optional tuple`(subsystem: String, category: String)?`, but in ObjC we have to manually check this.
If both are nil, we use the default log. Otherwise we create one using `os_log_create`.

Classes/DDOSLogger.h
```diff
@@ -36,3 +36,9 @@ API_AVAILABLE(macos(10.12), ios(10.0), watchos(3.0), tvos(10.0))
 @property (class, readonly, strong) DDOSLogger *sharedInstance;
 
 @end
+
+@interface DDOSLogger (Variations)
+- (instancetype)withSubsystem:(NSString *)string;
```
ffried: With the initializer, I'd remove these methods. They only cause confusion. Will they return the same instance? Will they create a new one (copy)?
I think most people will use this logger like the file logger: Create one (with a subsystem / category combination) and then add it.

lolgear:
@ffried 
I have checked it.
`os_log_create()` can't handle nil parameter in neither subsystem nor category.

Classes/DDOSLogger.m
```diff
@@ -79,4 +85,30 @@ - (DDLoggerName)loggerName {
...
+    if (_logger == nil)  {
+        _logger = [self getLogger];
+    }
+    return _logger;
+}
+@end
+
+@implementation DDOSLogger (Variations)
+- (instancetype)initWithSubsystem:(NSString *)subsystem category:(NSString *)category {
+    NSAssert(subsystem != nil || category != nil, @"Neither subsytem nor category can be nil. Otherwise, OSLogger falls back to OS_LOG_DEFAULT.");
```
ffried: This is not completely correct. I can pass nil for both subsystem and category and should get back a valid logger.
```suggestion
    NSAssert((subsystem == nil) == (category == nil), @"Either both subsytem and category or neither can be nil.");
```

ffried: Also, we should probably override the default initializer (`- init`) and forward it to this one with both parameters being nil.

Classes/DDOSLogger.m
```diff
@@ -17,6 +17,12 @@
 
 #import <os/log.h>
 
+@interface DDOSLogger ()
+@property (copy, nonatomic, readwrite) NSString *subsystem;
```
ffried: These don't have to be `readwrite` anymore. They're readonly.

Classes/DDOSLogger.h
```diff
@@ -36,3 +36,7 @@ API_AVAILABLE(macos(10.12), ios(10.0), watchos(3.0), tvos(10.0))
 @property (class, readonly, strong) DDOSLogger *sharedInstance;
 
 @end
+
+@interface DDOSLogger (Variations)
+- (instancetype)initWithSubsystem:(NSString *)subsystem category:(NSString *)category;
```
ffried: Since this should become the designated initializer, it can also go into the main `@interface`.

Classes/DDOSLogger.m
lolgear: @ffried Do you mean ivars or logger initialization in designed initializer?

ffried: @lolgear ivars. I just checked, DDFileLogger also uses ivars in the initializer.

Classes/DDOSLogger.m
```diff
@@ -17,10 +17,49 @@
...
+        _subsystem = [subsystem copy];
+        _category = [category copy];
+    }
+    return self;
+}
+
 static DDOSLogger *sharedInstance;
 
+- (instancetype)init {
+    if (sharedInstance != nil) {
```
ffried: Why should it be wrong to create another DDOSLogger if we have a shared instance? The other loggers also allow multiple instances besides the shared instance.

lolgear: @ffried 
I suppose that you suggest to completely remove `sharedInstance`?
But even `DDTTYLogger` has the same logic in `-init`.

ffried: `sharedInstance` is fine by me for those who want to use it. But we shouldn't ban those that don't want to use it.
I'm usually comparing with DDFileLogger, since this is currently the only logger that isn't deprecated 😉

lolgear: @ffried 
`DDOSLogger` already has `sharedInstance` of `OS_LOG_DEFAULT` inside. We have correct lookup in `DDOSLogger` for system `os_log` no matter how to store `DDOSLogger`. Maybe just remove this `sharedInstance`?

ffried: We could. But this would be a breaking change.

ffried: For now this is fine, I think.

Issue #1048, opened 2019-03-03
`-[DDFileLogger currentLogFileInfo]` will always create a new info when `doNotReuseLogFiles=YES`


Whirlwind:
I set the `doNotReuseLogFiles` to `YES`, and I hope the `DDFileLogger` create the new log file when I init the `DDFileLogger`. But before I release the `DDFileLogger`, I will get the `currentLogFileInfo` many times. I will get many infos!!  I check the master branch:

```objective-c
- (DDLogFileInfo *)lt_currentLogFileInfo {

    // Check if the file we've found is still valid. Otherwise create a new one.
    if (newCurrentLogFile != nil && [self lt_shouldUseLogFile:newCurrentLogFile]) {
        if (isResuming) {
            NSLogVerbose(@"DDFileLogger: Resuming logging with file %@", newCurrentLogFile.fileName);
        }
        _currentLogFileInfo = newCurrentLogFile;
    } else {
        NSString *currentLogFilePath = [_logFileManager createNewLogFile];
        _currentLogFileInfo = [[DDLogFileInfo alloc] initWithFilePath:currentLogFilePath];
    }
}

- (BOOL)lt_shouldUseLogFile:(nonnull DDLogFileInfo *)logFileInfo {
    NSAssert([self isOnInternalLoggerQueue], @"lt_ methods should be on logger queue.");
    NSParameterAssert(logFileInfo);

    if (logFileInfo.isArchived) { // Archived log files are no longer valid for use.
        return NO;
    }

    if (_doNotReuseLogFiles || [self lt_shouldLogFileBeArchived:logFileInfo]) { // This log file needs to be archived. No longer valid.
        logFileInfo.isArchived = YES;
        NSString *archivedLogFilePath = [logFileInfo.fileName copy];

        if ([_logFileManager respondsToSelector:@selector(didArchiveLogFile:)]) {
            dispatch_async(_completionQueue, ^{
                [self->_logFileManager didArchiveLogFile:archivedLogFilePath];
            });
        }

        return NO;
    }

    // All checks have passed. It's valid.
    return YES;
}
```
The `[self lt_shouldUseLogFile:newCurrentLogFile]` always be `NO`, because the `_doNotReuseLogFiles` is `YES`. So it will call `createNewLogFile` everytime!

I think that `currentLogFileInfo` should be a getter mathod, and it should not be a handler method! Right?

ffried:
Thanks for reporting this.
[...]

PR #1049, opened 2019-03-03
Fix creation of too many log files

ffried:

This merge request fixes / refers to the following issues: #1048 

### Pull Request Description

When `doNotReuseLogFiles` was set to yes, the `DDFileLogger` would create a new log file on every call to `currentLogFileInfo`. This is fixed now, since we only archive a log file if we're resuming.

### DIFF
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -1018,7 +1018,7 @@ unsigned long long const kDDDefaultLogFilesDiskQuota   = 20 * 1024 * 1024; // 20
     }
 
     // Check if the file we've found is still valid. Otherwise create a new one.
-    if (newCurrentLogFile != nil && [self lt_shouldUseLogFile:newCurrentLogFile]) {
+    if (newCurrentLogFile != nil && [self lt_shouldUseLogFile:newCurrentLogFile isResuming:isResuming]) {
         if (isResuming) {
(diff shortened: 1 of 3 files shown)


Classes/DDFileLogger.m
```diff
@@ -1031,15 +1031,17 @@ - (DDLogFileInfo *)lt_currentLogFileInfo {
     return _currentLogFileInfo;
 }
 
-- (BOOL)lt_shouldUseLogFile:(nonnull DDLogFileInfo *)logFileInfo {
+- (BOOL)lt_shouldUseLogFile:(nonnull DDLogFileInfo *)logFileInfo isResuming:(BOOL)isResuming {
```
lolgear: Suppose that somebody is using this method ( for some reason ). His implementation will fail due to  unrecognized selector or even worse - he didn't know why his overriding method doesn't change anything. So, his frustration will arise.

Is it intended behavior in hot fix?

I suggest to either forward invocation to this method from old one ( with resuming parameter equal to previous computation scheme of parameter `isResuming` ) or add explicit assertion or NSLogInfo/NSLogWarning to notify developer about new API.

ffried: `lt_shouldUseLogFile:` was before and is still private API (as are all `lt_`-APIs). If somebody relied on this (which I highly doubt, since it's been in `master` for only a few days), they must expect it to break at any time.

sushichop:
@ffried 
Nice!

PR #1053, opened 2019-03-14
Sync internal queues to prevent cleaning up log files too soon in tests

ffried:

### Pull Request Description

On the iOS simulator, we would have a race condition where test files would be cleaned up before the tested file logger was able to clean up after itself. This fixes this by syncing DDLog's queue and the file logger's queue before deleting the log files/.

### DIFF
--- a/Tests/Tests/DDFileLoggerTests.m
+++ b/Tests/Tests/DDFileLoggerTests.m
@@ -40,7 +40,9 @@ static const DDLogLevel ddLogLevel = DDLogLevelAll;
     
     [DDLog removeAllLoggers];
     // We need to sync all involved queues to wait for the post-removal processing of the logger to finish before deleting the files.
+    NSAssert(![self->logger isOnGlobalLoggingQueue], @"Trouble ahead!");
     dispatch_sync([DDLog loggingQueue], ^{
+        NSAssert(![self->logger isOnInternalLoggerQueue], @"Trouble ahead!");
         dispatch_sync(self->logger.loggerQueue, ^{
             /* noop */
         });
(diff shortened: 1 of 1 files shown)


Tests/Tests/DDFileLoggerTests.m
```diff
@@ -39,6 +39,12 @@ - (void)tearDown {
     [super tearDown];
     
     [DDLog removeAllLoggers];
+    // We need to sync all involved queues to wait for the post-removal processing of the logger to finish before deleting the files.
+    dispatch_sync([DDLog loggingQueue], ^{
```
lolgear: Maybe we should add check for distinct queues?
Because double syncing isn't safe on one queue.
Maybe we should add check for distinct queues?
Because double syncing isn't safe on one queue.

Something like
```objective-c
if (queue != self->logger.loggerQueue) {/*inner sync*/}
```

ffried: To be fair, if this should ever happen, half of CocoaLumberjack will deadlock 😉

lolgear: @ffried Could you leave a comment about it in this double-sync block? :)

ffried: I can, but do we need it? There are like hundreds of places in CocoaLumberjack where we first sync the DDLog queue and then the logger's queue.
e.g. `- [DDAbstractLogger logFormatter]`

ffried: `- [DDFileLogger flush]`
`- [DDFileLogger rollingFrequency]`
`- [DDFileLogger maximumFileSize]`
...

ffried: Oh well, I've added asserts... ¯\_(ツ)_/¯

sushichop:
@ffried 
Great!

Issue #1050, opened 2019-03-13
Security enhancements


stonedauwg:
There are a few instances of This project using so-called banned C standard APIs, banned from an OWASP perspective. For example strlen, strncpy etc. Can these be replaced with security friendly alternatives? [Here is a list](https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/master/StaticAnalyzer/views/ios/binary_analysis.py) of some code smells looked for during binary analysis. Running a scan on a app that uses CocoaLumberjack flags some of these "bad" C API calls

lolgear:
@stonedauwg 
Could you show an actual result of testing which contains a list of C API calls that smell?

stonedauwg:
Sure thing. A test with the following lines of code:

`
[DDLog addLogger:[DDOSLogger sharedInstance]]; // Uses os_log
DDFileLogger *fileLogger = [[DDFileLogger alloc] init]; // File Logger
fileLogger.rollingFrequency = 60 * 60 * 24; // 24 hour rolling
fileLogger.logFileManager.maximumNumberOfLogFiles = 7;
[DDLog addLogger:fileLogger];
`
produced this "banned API" report from MobSF

Binary make use of banned API(s)
Insecure The binary may contain the following banned API(s) _strlen.

Binary make use of malloc Function
Insecure The binary may use malloc function instead of calloc

Note that I had to include the actual source for CL instad of use it as a carthage built framework to get it to be flagged for anything. That may be as result of the binary getting further obfuscated when using as a framework. MobSF does a simple string search for static analysis on the iOS binary so is sensative to that (see link in my prev post)

The results do make sense as a simple search on this project's github repo shows use of C API calls included in the "banned" list like strlen, malloc etc

PR #1054, opened 2019-03-15
One warning fix

hhanesand:

PR #1056, opened 2019-03-19
Replace banned APIs with safe ones

ffried:

This merge request fixes / refers to the following issues: #1050 

### Pull Request Description

This replaces the banned APIs in the list mentioned in #1050 with their safe counterparts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## [3.5.3 - Xcode 10.1 on ??, 2019](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.5.3)
+
+### Public
+- ...
+
+### Internal
+- Remove banned APIs (#1056)
+
 ## [3.5.2 - Xcode 10.1 on 15th, 2019](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.5.2)
 
 ### Public
--- a/Classes/DDLog.m
+++ b/Classes/DDLog.m
@@ -619,7 +619,7 @@ static NSUInteger _numProcessors;
 
         NSUInteger bufferSize = numClasses;
 
-        classes = numClasses ? (Class *)malloc(sizeof(Class) * bufferSize) : NULL;
+        classes = numClasses ? (Class *)calloc(bufferSize, sizeof(Class)) : NULL;
         if (classes == NULL) {
             return @[]; //no memory or classes?
         }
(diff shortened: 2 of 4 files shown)


lolgear:
@ffried 
Could you merge #1054 first? ( If it is necessary ).

ffried:
@lolgear done

lolgear:
I neither like 1 as a length of desired string nor 2 as a max length of string in `strnlen`.
Maybe it is better to use macros here or const/static?
I don't know, it is weird :/

ffried:
What's wrong with these values? 1 as length is simply `malloc(size * 1)`. And since `strnlen` will have to check for `\0` in the string, 2 as max len will make sure it doesn't look too far but gives up whenever the string is any longer than 1. Since this is an arbitrary choice, I've added the comment there.

The first is just "normal" IMHO. The second is arbitrary, but only for this specific location, so macros or static const won't do any good either. That's what comments are for.

OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lolgear:
@ffried 
Fine, right.

I think about "pure" solution here.
1. Check first symbol of string for not to be null.
2. Check first symbol to be '/'.
3. Check second symbol to be null.


```c
// If I am correct about comparison here. Not Sure
if (name != '\0' && name == '/' && *(++name) == '\0') {}
```

lolgear:
@ffried 
Ah!
I see, you `calloc` 1s instead of `sizeof(char)` :)

ffried:
@lolgear You're actually right. We could express these as `(char *)calloc(_appLen + 1, sizeof(char))`... That sounds better indeed. Let's change that.

lolgear:
@ffried 
What about C pointer arithmetics?
Do you admire it? :)

ffried:
@lolgear What about it? 🙂

lolgear:
@ffried 

```c
// If I am correct about comparison here. Not Sure
if (name != '\0' && name == '/' && *(++name) == '\0') {}
```

ffried:
@lolgear I think it would be correct. But from a readability perspective much worse than `strnlen(name, 2) == 1` 😉

lolgear:
@ffried 
So, let's merge it?

ffried:
@lolgear Fine by me. Let's wait for the CI 🙂

sushichop:
LGTM👍
Now I found another banned API(function). So I'll create a PR for it later.

PR #1060, opened 2019-03-27
Improve DDAssert & DDAssertionFailure

diederich:

* [n/a] I have added the required tests to prove the fix/feature I am adding
* [n/a] I have updated the documentation (if necessary)

### Pull Request Description
This PR improves the newly added DDAssert and DDAssertionFailure:
* evaluate message only once to prevent side-effects
* always log the assertion message
* only call assertionFailure in DEBUG builds (fixes undefined behaviour)
  We shouldn't be calling assertionFailure in non-debug builds. Docs for assertionFailure say:
  ```
  /// In -Ounchecked builds, the optimizer may assume that this function is
  ///   never called. Failure to satisfy that assumption is a serious
  ///   programming error.
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### Public
 - Additional compatibility with Swift 5 (backwards compatible with Swift 4) (#1043)
+- improve DDAssert and DDAssertionFailure (#1060)
+  - evaluate message only once
+  - always log the assertion message
+  - fix undefined behaviour
 
 ### Internal
 - Remove banned APIs (#1056) (#1057)
--- a/Classes/DDAssert.swift
+++ b/Classes/DDAssert.swift
@@ -24,9 +24,9 @@
  *     The default is an empty string.
  */
 @inlinable
-public func DDAssert(_ condition: @autoclosure () -> Bool, _ message: @autoclosure () -> String = "", level: DDLogLevel = DDDefaultLogLevel, context: Int = 0, file: StaticString = #file, function: StaticString = #function, line: UInt = #line, tag: Any? = nil, asynchronous async: Bool = false, ddlog: DDLog = DDLog.sharedInstance) {
+public func DDAssert(_ condition: @autoclosure () -> Bool, _ message: @autoclosure () -> String = "", context: Int = 0, file: StaticString = #file, function: StaticString = #function, line: UInt = #line, tag: Any? = nil, asynchronous async: Bool = false, ddlog: DDLog = DDLog.sharedInstance) {
     if !condition() {
-        DDLogError(message(), level: level, context: context, file: file, function: function, line: line, tag: tag, asynchronous: async, ddlog: ddlog)
+        DDLogError(message(), level: .all, context: context, file: file, function: function, line: line, tag: tag, asynchronous: async, ddlog: ddlog)
         Swift.assertionFailure(message(), file: file, line: line)
     }
 }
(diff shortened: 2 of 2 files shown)


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ffried:
@diederich `-Ounchecked` and release builds are not the same. Release builds are simply `-O` (debug would be `-Onone`). It's totally fine to call `assertionFailure` in release builds, as stated in the header docs:
> * In -O builds, has no effect.

ffried:
IMHO we should leave the assertionFailure call as is. Basically, `DDAssert` and `DDAssertionFailure` are as unsafe to call in `-Ounchecked` builds as is `assertionFailure` in Swift.

diederich:
> @diederich -Ounchecked and release builds are not the same. Release builds are simply -O (debug would be -Onone). It's totally fine to call assertionFailure in release builds, as stated in the header docs

True, the correct check would be for -Onone, but nothing like that exist I think.
I think the idea of those assertions is call them in all cases, even in -Ounchecked, so that an error gets logged. 
Asking the other way around - do you think it can hurt to disable the calls in "Release" builds?
Maybe add a clarifying comment?

diederich:
Another option I considered was removing them completely, as I wasn't sure if there can actually be a 'correct' implementation. Not being allowed to call them from client code, or better, depending on the build-setting, kind of defeats the use case in the first place :-/

ffried:
The problem is, that `DEBUG`  is only a define, no guarantee. So it's totally possible to run a release build with `DEBUG` defined or the other way round: run a debug build without it. This is one reason why e.g. [SwiftNIO uses `assert`])(https://github.com/apple/swift-nio/blob/9b78557affaf4f600b0c44588ea86143d3b1451b/Sources/NIO/Utilities.swift#L23) for running code only in debug builds.

ffried:
Like I said, I think the `DDAssert`-functions are just a wrapper around `Swift.assert` that perform additional logging. But apart from that, they should be treated the same as `Swift.assert`. There are a few more points that speak for this:
- `-Ounchecked` builds aren't really "recommended".
- If I was still using a `-Ounchecked` build, I'm looking for performance. In this case, I probably don't wan't any logging either. So I'll probably guard my calls to any logging functions anyways.

diederich:
I did check if we set DEBUG in the project file, so I though it's in our hands at least.

Anyways, I'm happy to remove them altogether, given they're not usable how we thought they would be :-) (cc @ole)

ffried:
A more detailed discussion about this can be found [here](https://forums.swift.org/t/support-debug-only-code/11037/31) btw.

diederich:
> Like I said, I think the DDAssert-functions are just a wrapper around Swift.assert that perform additional logging.

Yes, agreed. With that perspective they still make sense and should stay. 

(They were copied from our Utils library with a different use case in mind - but I guess that doesn't really matter anymore, and we can easily add the checks on our side.)

Let me get rid of those #ifdef's - :pray: for the review!

ffried:
@diederich Also, I think the `level` parameter should stay. Thinking more about this, does it make sense to log assertion failures in release builds? Or should we opt for a similar pattern than in NIO where we call the log-function in a closure passed to `assert`?
We could further introduce `DDPrecondition` and `DDPreconditionFailure` that do the same with `Swift.precondition` and `Swift.preconditionFailure` giving the option to also crash in `-O` builds.

diederich:
Our use case is to just log errors in release builds, even when assertions are disabled. Additionally, in case someone builds with an app with -Ounchecked, I want to avoid undefined behaviour.

Calling preconditionFailure means we'd get a crash-report (hopefully :-)) - so no need for additional logging in that case. (Now that you say that, I'm not sure if the message passed to preconditionFailure shows up in the logs though.. :-))

--- 

Let me try to put some scenarios together with the previous implementation / + the case of not logging:
a) if assertions are enabled and I'm in the debugger: it traps -> no need to log
b) if assertions are enabled and I'm not in the debugger: it crashes with a crash-report (and hopefully with the assertion logged in it. At least I'd see which assertion it was from the stacktrace)
c) if assertions are disabled and it's not logging: it's the same as calling plain swift assert
d) if assertions are disabled and it's built with with -Ounchecked: we're in undefined behaviour land

I think I might be still just thinking from our use case, so please correct / add some scenarios.

ffried:
@diederich I see your point in use case c).
So for now I think we should just leave the Swift stdlib calls in place and always log the error. But also keep the `level` parameter, since people might want to turn of logging completely.

diederich:
OK, cool. Thanks for the review!

Given we don't do c), I think we should just drop this PR then, as the early evaluation of the message might not be necessary iff callers can disable logging of the message, and should thus be left to DDLogError & Swift.assert to decide. 

It's fine to use DDAssert as it is instead of Swift.assert, if people want a log entry in `-O`-builds.
If someone's using -Ounchecked, they better know what they're doing anyways :-)

PR #1071, opened 2019-05-09
fix missing ios system framwork dependence

euanchan:

### Pull Request Description

While enable `use_frameworks` in Podfile, errors encountered while link the project cause CocoaLumberjack's podspec has leak of enough system framework dependence.

the error info:
```
Undefined symbols for architecture arm64:
  "_CGContextFillRect", referenced from:
      +[DDTTYLogger getRed:green:blue:fromColor:] in DDTTYLogger.o
  "_CGContextRelease", referenced from:
      +[DDTTYLogger getRed:green:blue:fromColor:] in DDTTYLogger.o
  "_CGColorSpaceRelease", referenced from:
      +[DDTTYLogger getRed:green:blue:fromColor:] in DDTTYLogger.o
  "_OBJC_CLASS_$_UIColor", referenced from:
      objc-class-ref in DDTTYLogger.o
  "_CGContextSetFillColorWithColor", referenced from:
      +[DDTTYLogger getRed:green:blue:fromColor:] in DDTTYLogger.o
  "_CGColorSpaceCreateDeviceRGB", referenced from:
      +[DDTTYLogger getRed:green:blue:fromColor:] in DDTTYLogger.o
  "_CGBitmapContextCreate", referenced from:
      +[DDTTYLogger getRed:green:blue:fromColor:] in DDTTYLogger.o
  "_UIApplicationWillTerminateNotification", referenced from:
      -[DDLog init] in DDLog.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

To fix this, add the missing frameworks dependence to podspec 
...

### DIFF
--- a/CocoaLumberjack.podspec
+++ b/CocoaLumberjack.podspec
@@ -30,6 +30,7 @@ Pod::Spec.new do |s|
   s.subspec 'Core' do |ss|
     ss.source_files         = 'Classes/CocoaLumberjack.h', 'Classes/DD*.{h,m}', 'Classes/Extensions/*.{h,m}', 'Classes/CLI/*.{h,m}'
     ss.private_header_files = 'Classes/DD*Internal.{h}'
+    ss.ios.frameworks       = [ "Foundation", "UIKit", "CoreGraphics"]
   end
 
   s.subspec 'Swift' do |ss|


ffried:
@euanchan Thanks for contributing this fix!
I'm not too familiar with CocoaPods, but I assume we need to do something similar for macOS as well, right? Also watchOS and tvOS, or does CocoaPods apply the iOS settings for those platforms?

ffried:
@CocoaLumberjack/collaborators We do have `use_frameworks` in our README.md...
Did we break something when we refactored / simplified the podspec a while back?

lolgear:
@euanchan 
Wow!
I am astonished by this fact!
I suppose that "Foundation", "UIKit" and "CoreGraphics" are default for iOS apps and Xcode silently adds them for you.
I can't imagine how you can build iOS app without any of them...

bpoplauschi:
@lolgear I had the same impression (that those frameworks are linked by default).

ffried:
Yea, I still wonder what's different here. I'm almost certain we have users out there using `use_framework`.

Xcode _should_ determine (system) dependencies automatically, so it should link those frameworks automatically.

ffried:
I've just checked. Even before the simplification of the podspec, we didn't specify any frameworks directly.

lolgear:
I would like to know build environment of this situation.

@euanchan 
Can you share versions of Xcode, deployment target, macOS, etc.?

euanchan:
Commonly I use cocoapods without `use_framework`, someone told me while they use my tools [MTHawkeye](https://github.com/meitu/MTHawkeye), he failed at building, and he use `use_frameworks`.

So I create an empty swift project to reproduce the situation, then found that both `FLEX` and `CocoaLumberjack` has leak some of the dependence.

euanchan:
@lolgear Xcode 10.2, macOS 10.14.3, and the deployment target to 8.0

OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">The CocoaPods podspec was modified but the Carthage project wasn’t. Did you forget to update the xcodeproj?</td>
    </tr>
  </tbody>
</table>

</p>

euanchan:
Sorry for missing some information for you, I use an template to create the project, and it enable ccache by default. I've ensure only when I disable the `CLANG_ENABLE_MODULES`, the error will raise.

what have I did in the pod file 

```
platform :ios, "8.0"

workspace 'TestPod'

use_frameworks!

target 'TestPod' do

  #pod 'FLEX'#, :path => '/Users/euanc/workspace/github/_devTools/FLEX', :configurations => ['Debug'], :inhibit_warnings => true

  #pod 'MTHawkeye', :path => '~/workspace/meitu/ios/MTHawkeye', :configurations => 'Debug', :inhibit_warnings => true

  pod 'CocoaLumberjack'

  post_install do |installer_representation|
    installer_representation.pods_project.targets.each do |target|
      target.build_configurations.each do |config|
        # trun off Enable Modules
        config.build_settings['CLANG_ENABLE_MODULES'] = 'NO'

        # 
        #config.build_settings['CC'] = '$(PODS_ROOT)/../ccache-clang'
      end
    end
  end

  target 'TestPodTests' do
  end

  target 'TestPodUITests' do
  end

end
```

ffried:
That makes sense now. `CLANG_ENABLE_MODULES` is the switch (or one of them) that turns on the auto-linking magic. Is there a particular reason for disabling it in your case?

euanchan:
Yep, for we're trying to reduce the time cost of compiling, we introduce ccache in our projects, it needs to close `CLANG_ENABLE_MODULES.

sushichop:
Thanks for your info. I understood you disabled clang moduels to use ccache because it does not support clang modules. It is correct, isn't it?
At first, `Foundation` is not necessary because CocoaPods links it during building as I mentioned above.

Then I think we have 2 options in order to achieve @euanchan 's  request.
cc: @CocoaLumberjack/collaborators 

1.  Updates podspec like this pull request. 
     In this case, CocoaLumberjack needs to be updated for other platforms(macOS, tvOS, and watchOS). and it maybe complicated. 
    But `CococaLumberjack` can supports the case like this time.

2. Add scripts at post_install in the Podfile to link system frameworks.
    In this case, CocoaLumberjack does not need to be updated and it is simply.
    But developers like this time need to add scripts at post_install.

Which one is preferable?

sushichop:
> 2\. Add **more scripts(but only 1 line)** at post_install in the Podfile to link system frameworks.
>     In this case, CocoaLumberjack does not need to be updated and it is simply.
>     But developers like this time need to add scripts at post_install.

I gave a shot about option 2. The sample Podfile is here. Is this enough or not?
https://gist.github.com/sushichop/838d27498cb5504a63145fb85213921b

```ruby
platform :ios, '8.0'

target 'SampleIOSApp' do
  use_frameworks!

  pod 'CocoaLumberjack'

  post_install do |installer|
    installer.pods_project.targets.each do |target|
      target.build_configurations.each do |config|
        config.build_settings['CLANG_ENABLE_MODULES'] = 'NO'
        # e.g. Sets Build Settings for using cclang cache.
        # config.build_settings['CC'] = '$(PODS_ROOT)/../ccache-clang'
      end

      target.add_system_framework(['CoreGraphics', 'UIKit'])
    end
  end

end
```

ffried:
I'm in favor of option 2. IMHO CocoaLumberjack should be configured for what is default and recommended by Apple (and therefore also what most projects will use).

lolgear:
@sushichop 
We should add an example of Podfile setup in Readme, right?

Also, does CocoaPods check uniqueness of system frameworks? It isn't a big deal, though.

sushichop:
Yeah, TBH, I'm also in favor of option 2. IMO, I think current podspec is still better. 
As far as I remember, clang module has been enabled as a default sinceXcode 6 or so.
If we considered the case clang module is disabled, we must need to consider what frameworks we should link whenever a New Xcode version has been released. 
This should be avoided.

sushichop:
> We should add an example of Podfile setup in Readme, right?

Thanks. I can actually add an example of Podfile about this case, but, at the same time, I think  `disabling clang module` and using `add_system_framework`n Podfile are rare special cases.
So I'm afraid that adding this example may cause confusion...😅

> Also, does CocoaPods check uniqueness of system frameworks? It isn't a big deal, though.

I don't know(but I assume it does not check it probably)...

sushichop:
@euanchan 

Then, can you use `add_system_framework` method in Podfile?
I believe that this solution can meet your needs.

Thanks!

ffried:
I'm with @sushichop on this one. Adding an example will only cause confusion. At best, we can add a documentation section somewhere about this. But I think people actively disabling `CLANG_ENABLE_MODULES` should also know how to deal with the side-effects of this. Adding `add_system_framework` to any potential Pods they depend on is one of them.

ffried:
I'm also not sure if CocoaPods should really care about this. CocoaPods is not a build setting validator. That's what Xcode (and the underlying tools) do. And it produced above's error stating what is wrong.

lolgear:
Nevertheless, we have an experience with it and should save it somewhere to prevent further questions.
If this PR is a nice place of knowledge, close it as solved.
Otherwise, add "documentation notes".

bpoplauschi:
I agree with @ffried

euanchan:
@sushichop That would work, I'll add `add_system_framework` by myself. Thank you guys.

stagedataloss1:
test123test123

PierreBrisorgueil:
Hey @sushichop 

Hello, I got the error today with SPM. I didn't understand why; I looked for CLANG_ENABLE_MODULES in my workspace and found that some packages had it expressly set to YES. I added the config on CocoaLumberjack, and this fixed it. It was a fresh install if that helps some ...

Screenshot 2021-03-17 at 22 32 37

sushichop:
@PierreBrisorgueil 

I'm not sure but some xcconfig or xcode's project may have disabled CLANG_ENABLE_MODULES...
Thanks for your information.

PR #1076, opened 2019-06-17
Expose current log file info in `didLog` methods.

hhanesand:
I removed a public variable to provide queue protection in d0fed54ea39d710918f5b9d3627d32a50f889c46 .

### DIFF
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -1135,6 +1135,14 @@ static int exception_count = 0;
 }
 
 - (void)didLogMessage {
+
+}
+
+- (void)willLogMessage:(DDLogFileInfo *)logFileInfo {
+
+}
+
+- (void)didLogMessage:(DDLogFileInfo *)logFileInfo {
     [self lt_maybeRollLogFileDueToSize];
 }
 
@@ -1217,13 +1225,23 @@ static int exception_count = 0;
     }
 
     @try {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
         [self willLogMessage];
+#pragma clang diagnostic pop
+
+        [self willLogMessage:_currentLogFileInfo];
 
         NSFileHandle *handle = [self lt_currentLogFileHandle];
         [handle seekToEndOfFile];
         [handle writeData:data];
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
         [self didLogMessage];
+#pragma clang diagnostic pop
+
+        [self didLogMessage:_currentLogFileInfo];
     } @catch (NSException *exception) {
         exception_count++;
 
--- a/Classes/DDFileLogger.m
+++ b/Classes/DDFileLogger.m
@@ -1223,7 +1223,11 @@ static int exception_count = 0;
     return [super methodSignatureForSelector:aSelector];
 }
 
-- (void)forwardInvocation:(NSInvocation *)anInvocation {}
+- (void)forwardInvocation:(NSInvocation *)anInvocation {
+    if (anInvocation.selector != @selector(dummyMethod)) {
+        [super forwardInvocation:anInvocation];
+    }
+}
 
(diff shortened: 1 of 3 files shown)


Classes/DDFileLogger.m
```diff
@@ -1217,13 +1225,23 @@ - (void)lt_logData:(NSData *)data {
     }
 
     @try {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
         [self willLogMessage];
```
ffried: We should only call this, if the new version is not implemented.

Classes/DDFileLogger.m
```diff
@@ -1217,13 +1225,23 @@ - (void)lt_logData:(NSData *)data {
...
+
+        [self willLogMessage:_currentLogFileInfo];
 
         NSFileHandle *handle = [self lt_currentLogFileHandle];
         [handle seekToEndOfFile];
         [handle writeData:data];
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
         [self didLogMessage];
```
ffried: same here

OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

hhanesand:
This would crash if subclasses called the deprecated methods since we were no longer implementing it. Fixed by forwarding to `dummyMethod`. I've checked that `respondsToSelector:` still returns `NO` even though we are forwarding the methods.

hhanesand:
@ffried

ffried:
@hhanesand Hmm, I'm not sure if this isn't causing more confusion than it should. When I wrote my comment, I didn't consider the fact, that we need to implement it ourselves, too.

Couldn't we just implement the two methods (empty) and add `[self class] != [DDFileLogger class]` to the `respondsToSelector:` checks?
So that `implementsDeprecated{Will,Did}Log` is only true, if we're being subclassed.

hhanesand:
I don't think that would work?

If we are subclassed, and we implement the deprecated methods, then `respondsToSelector:` will always return `YES`. We'd need a way to test if the subclass overrides the method. Changing the check to  `respondsToSelector: && [self class] != [DDFileLogger]` would mean we always call the deprecated method, regardless of the subclass's overriding state. Maybe I'm not understanding.

This way, we don't implement the deprecated methods, so `repondsToSelector:` performs the "does the subclass implement method" check. Furthermore, to avoid crashes, we forward any `[super will/didLogMessage]` calls that may or may not exist into another method.

hhanesand:
@ffried

ffried:
@hhanesand You're absolutely right. Not my best day today, sorry!
There are ways to get the methods of a subclass, but that would definitively be more complicated than what you've done now.

Just to jumpstart my slow brain today: Why do we need to override `forwardInvocation:`?

hhanesand:
`[NSObject forwardInvocation:]` docs :
```
To respond to methods that your object does not itself recognize, you must override methodSignatureForSelector: in addition to forwardInvocation:. The mechanism for forwarding messages uses information obtained from methodSignatureForSelector: to create the NSInvocation object to be forwarded. Your overriding method must provide an appropriate method signature for the given selector, either by pre formulating one or by asking another object for one.
```

You can see it being used in ``DDFileLogger+Buffering.m` as well. We may not actually need to override `forwardInvocation:`, haven't tested.

ffried:
@hhanesand Ah right, I see. Should we check the `invocation.selector` and `invocation.target` if it's our dummy method and otherwise invoke `[super forwardInvocation:invocation]`?
At least in the sample in the docs, they do something similar.

hhanesand:
Yeah, that'd probably be wise.

ffried:
Great. Thanks for your patience today with me 😀
LGTM now. 👍

hhanesand:
No problem! Happens to all of us :)

Issue #1080, opened 2019-07-03
Could not disable timestamp


Whirlwind:
I use the DDTTYLogger:

```
class MyFormatter: DDLogFormatter {
    func format(message logMessage: DDLogMessage) -> String? {
        if logMessage.message == "\n" { return logMessage.message }
        return logMessage.message.split(separator: "\n").map { ">> \($0)" }.joined(separetor: "\n")
    }
}
let console = DDTTYLogger.sharedInstance!
console.logFormatter = MyFormatter()
logger.add(console, with: .info)

logger.log("A")
logger.log("\n")
logger.log("B")
```
I get the output:
```
>> A
2019-07-03 22:11:09:508 XX[34846:4138957]

>> B
```

The timestamp could not be disable, the `DDTTYLogger.m` said:
```
- (void)logMessage:(DDLogMessage *)logMessage {
    NSString *logMsg = logMessage->_message;
    BOOL isFormatted = NO;

    if (_logFormatter) {
        logMsg = [_logFormatter formatLogMessage:logMessage];
        isFormatted = logMsg != logMessage->_message;
    }
    ....
    if (isFormatted) {
        // Will not add the timestamp
   } else {
      // Will add the timestamp
   }
```
My formatter will return origin string when the message is `\n`, it is very bad. I think we could have a variable to disable the timestamp instead of check the `isFormatted`.

cyg02032015:
So.... How do i remove it ????
```
2020-02-19 18:59:39.712979+0800 xxxxxxxxx[21239:2116064] my print format
2020-02-19 19:01:37.939565+0800 xxxxxxxxx[21239:2116064] my print format
2020-02-19 19:01:37.939896+0800 xxxxxxxxx[21239:2116064] my print format
2020-02-19 19:47:41.476513+0800 xxxxxxxxx[21239:2116065] my print format
```

PR #1081, opened 2019-07-04
Add `useStandardStyleForCustomFormatters` property in DDTTYLogger, it…

Whirlwind:
I add a property `useStandardStyleForCustomFormatters` in DDTTYLogger, It will allow user apply the NSLogger Style to the custom logger formatter. And it will fix the #1080, too.

### DIFF
--- a/Classes/DDTTYLogger.h
+++ b/Classes/DDTTYLogger.h
@@ -104,6 +104,12 @@
 @property (nonatomic, readwrite, assign) BOOL automaticallyAppendNewlineForCustomFormatters;
 
 /**
+ * Apply standard NSLog style formatting when use custom formatter,
+ * It will insert timestamp, thread id and so on. Default value is NO.
+ **/
+@property (nonatomic, assign) BOOL useStandardStyleForCustomFormatters;
+
+/**
  * The default color set (foregroundColor, backgroundColor) is:
  *
  * - DDLogFlagError   = (red, nil)
--- a/Classes/DDTTYLogger.m
+++ b/Classes/DDTTYLogger.m
@@ -884,6 +884,7 @@ static DDTTYLogger *sharedInstance;
         _colorProfilesDict = [[NSMutableDictionary alloc] initWithCapacity:8];
 
         _automaticallyAppendNewlineForCustomFormatters = YES;
+        _useStandardStyleForCustomFormatters = NO;
     }
 
     return self;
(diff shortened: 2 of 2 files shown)


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lolgear:
@Whirlwind Could you add tests for this feature?

Whirlwind:
> What benefit do we get from setting this new property to YES?

If I have a formatter, and I still hope there are the timestamp at the beginning of every log, the property will allow use both formatter and timestamp.

ffried:
> If I have a formatter, and I still hope there are the timestamp at the beginning of every log, the property will allow use both formatter and timestamp.

Ok, I see that point. Although I'm not sure that this is the case very often. In theory, `DDTTYLogger` is doing to much work here. Formatting messages should always be the responsibility of a log formatter. IMHO there should be a default log formatter for DDTTYLogger.

@Whirlwind Do you think you could extract the formatting logic from DDTTYLogger into a `DDNSLogFormatter`? Since it uses the format used by `NSLog`, other loggers might also benefit from this (e.g. file loggers). For backwards compatibility, you could simply create an instance of this by default in `DDTTYLogger`. As soon as someone assigns its own log formatter, the custom formatting is gone.

Whirlwind:
Sorry, I have a try. But I could not understand the `iovec` and `writev(STDERR_FILENO, v, 13);` in the `DDTTYLogger.m`. Why not make a string with the `thread id`, `app name`, `timestamp`? Is it faster?

ffried:
@Whirlwind The "gather" variant of write (`writev`) has the advantage that the string does not need to be one contiguous buffer of characters. Instead they can be scattered in various locations in memory. While in theory a `NSString` would work, too, it would need to allocate the memory contiguously.

That being said, a log formatter only returns a normal `NSString`. So if you extract that particular logic into a `DDNSLogFormatter`, you'll need to switch to a normal `NSString` anyway. IMHO this is an acceptable tradeoff for having a cleaner `DDTTYLogger`. 🙂

ffried:
@Whirlwind any news on this?

PR #1082, opened 2019-07-04
add `DDLogSTD` to allow set the pipe for `DDTTYLogger`

Whirlwind:
The TTYLogger will always print the log to STDERR, I want to handle the pipe for the every log message.

The `std` paramter will only for TTY, so I do not add it to the `log` in the DDLog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Public
 - New `willLogMessage:` and `didLogMessage:` methods on `DDFileLogger` which provide access to the current log file info.
+- New `DDLogSTD` enum and `std` property on `DDTTYLogger` which provide set the output pipe to `STDERR` or `STDOUT` for every log message.
 
 ## [3.5.3 - Xcode 10.2 on Apr 24th, 2019](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.5.3)
 
--- a/Classes/CocoaLumberjack.swift
+++ b/Classes/CocoaLumberjack.swift
@@ -81,6 +81,7 @@ public func _DDLogMessage(_ message: @autoclosure () -> String,
                           function: StaticString,
                           line: UInt,
                           tag: Any?,
+                          std: DDLogSTD = .auto,
                           asynchronous: Bool,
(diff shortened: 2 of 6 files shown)


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lolgear:
@Whirlwind Could you add tests for this feature?

Whirlwind:
Sorry, I not find the test in the project. Emmm..., it is in other project.  T_T

ffried:
> Sorry, I not find the test in the project. Emmm..., it is in other project. T_T
@Whirlwind Yea, tests are in a separate Xcode project

Whirlwind:
> Switch on the level and log errors to STDERR and all other messages to STDOUT

I did that. But I get some bad cases: My formatter will transform all error to other style, it will cause all log which print to STDERR become the style. It is bad. So the STDERR should be different with DDLevelError.

>  I'd rather have this parameter on TTYLogger itself. 

I did that, too. But it will change all logs to a pipe: STDOUT or STDERR, I hope that I could print a log to the STDERR, and others to the STDOUT.

ffried:
@Whirlwind I presume that you're doing a little to much work in your log formatters 😉 
To me it sounds like you're doing more than just _formatting_ your messages.

IMHO it sounds like the solution to your problem is a custom logger. The logic that decides whether something goes to STDERR or STDOUT should then be put into this logger. The formatter is then back to what it's supposed to do: formatting messages.
Or, if that's not possible for some reason, you could use associated objects to assign custom properties to a `DDLogMessage` in your formatter and use that as a basis for deciding which file no to log your message to in your custom logger (which could be a subclass of `DDTYLogger`).

What could make sense the, is to refactor `DDTTYLogger` a little bit to extract the call that does the writing into a separate method so that you can override it in your subclass and use a different file no there.

Whirlwind:
I think you are right. I will think about it and have a try. Thanks.

PR #1083, opened 2019-07-04
Swift package support.

lolgear:

#### Integration
- [x] Add Package.swift
- [x] Move headers to Include folder.
- [x] Add helper target for tests target ( Tests/Library ).
- [x] Add Tests targets to package.
- [x] Fix Xcode project.
- [x] Fix CocoaPods.
- [x] Fix tests targets.

#### Cleanup
- [ ] Cleanup DDLog (?)


lolgear:
This PR reveals several issues about imports and about structure of project.

DDLog.h contains following:

```
@class DDLogMessage;
@class DDLoggerInformation;

@protocol DDLogger;
@protocol DDLogFormatter;

@interface DDLog : NSObject

@protocol DDLogger <NSObject>
@protocol DDLogFormatter <NSObject>
@protocol DDRegisteredDynamicLogging

@interface DDLogMessage : NSObject <NSCopying>
@interface DDAbstractLogger : NSObject <DDLogger>
@interface DDLoggerInformation : NSObject
```

As we can see, it is a blob of various definitions and interfaces.

I suggest to cleanup DDLog (and move protocols and Data structures) in separate PR.

OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

lolgear:
Also, it seems that only relative paths are permitted.

In case of `Tests` target it means that all helpers headers ( or samples / scaffolds / mockings ) should either

1. be in the same source directory ( Tests ) 
2. be in separate target `CocoaLumberjackTestsHelper` or `CocoaLumberjackTestsTool` 
3. their imports should have relative paths `#import "../Library/File.h"`

lolgear:
@CocoaLumberjack/collaborators 
Can anybody review?

Package.swift
```diff
@@ -0,0 +1,35 @@
...
+
+import PackageDescription
+
+let package = Package(
+    name: "CocoaLumberjack",
+    products: [
+        // Products define the executables and libraries produced by a package, and make them visible to other packages.
+        .library(
+            name: "CocoaLumberjack",
+            type: .dynamic,
```
diederich: CocoaLumberjack also supports static linking I think. Could remove this and let clients decide?

Package.swift
```diff
@@ -0,0 +1,35 @@
...
+            targets: ["CocoaLumberjack"]),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
+        .target(name: "CocoaLumberjack",
```
diederich: Would there be a second target for the swift parts?

diederich:
Thanks for starting this @lolgear 

> Also, it seems that only relative paths are permitted.

Xcode 10.2 introduced a warning for framework headers which use quote-includes (https://developer.apple.com/documentation/xcode_release_notes/xcode_10_2_release_notes?language=objc). So I think public headers should actually use the  `#include <CococaLumberjack/...>` style.

Package.swift
lolgear: If `Swift Package manager` supports static linking, then, we should add static library also.

diederich: > /// A library's product can either be statically or dynamically linked. It
> /// is recommended to not declare the type of library explicitly to let the
> /// Swift Package Manager choose between static or dynamic linking depending
> /// on the consumer of the package.

https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescription.md
Removing it should be fine?

Package.swift
lolgear: Well, I am not sure how Swift Package manager solves dependencies for mixed targets. Moreover, I suppose that it doesn't support pure Objective-C targets, right?

diederich: I don't think it supports mixed (Objective-C/Swift) targets right now. So having an Objective-C and a swift target would make sense.

lolgear:
@diederich 
Well, I commented above about headers and Swift Package.
It seems that it can't handle angle-brackets for imports.
I think that we can solve this by adding `#ifdef`-s to check if Swift package manager build library.

Package.swift
lolgear: @diederich It doesn't. I filtered out swift files.

diederich:
> @diederich
> Well, I commented above about headers and Swift Package.
> It seems that it can't handle angle-brackets for imports.
> I think that we can solve this by adding `#ifdef`-s to check if Swift package manager build library.

Can you clarify where it doesn't work / can't be handled?

lolgear:
@diederich 

1. Open Package.swift in Xcode.
2. Try to build project.

Error: `'#import <CocoaLumberjack/DDLog.h>' file not found`

Package.swift
diederich: The Xcode project and the cocoa pods file seem to have 2 targets. One is "core", the other "CocoaLumberjackSwift" 
My guess would have been that Package.swift would also export those 2 options (and thus have those 2 targets)
So maybe add another target, which depends on this first one and includes the swift files? (I guess this could either be handled by excluding non swift files, or creating a separate directory for swift parts)

lolgear:
```objective-c
#ifndef DD_STR
#define DD_STR(str) #str
#endif

#ifndef DD_UNICODE_SYMBOL_BY_CODE
#define DD_UNICODE_SYMBOL_BY_CODE(code) CHAR(code)
#endif

#ifndef DD_SWIFT_PACKAGE_SUPPORT
#define DD_SWIFT_PACKAGE_SUPPORT(header_statement)
#if SWIFT_PACKAGE
#import header_statement
#else
#import DD_STR(<)DD_STR(CocoaLumberjack/STR(header_statement))DD_STR(>)
#endif
#endif
```

Or

```objective-c
#if SWIFT_PACKAGE
#import "DDLegacyMacros.h"
#import "DDLoggerNames.h"
#else
#import <CocoaLumberjack/DDLegacyMacros.h>
#import <CocoaLumberjack/DDLoggerNames.h>
#endif
```

Package.swift
lolgear: @diederich 
Consider latest commit. 
I comment out all stuff related to additional target in case of "unknown symbol DDLogLevel and its friends".

diederich:
We might get along by "massaging" the source layout a bit instead of #ifdef'ing in different build configurations.
This is a layout that seems to work without the include changes:
Screenshot 2019-07-08 at 22 48 35

Idea is to have one `include/CocoaLumberjack/` folder with all the public headers This kept the CocoaLumberjack/CocoaLumberjack.h  umbrella header, swift-pm can use it. (private headers could stay in `Classes` folder`). Maybe this could be simplified even more?

lolgear:
I suppose that you can swap `CocoaLumberjack/` and `include/`. Maybe you can even rename `Classes/` to achieve similar results.

lolgear:
@diederich Where do you put `CocoaLumberjackSwift.h`?
Its include statements seems to be broken. :/

"Import of the module on the same level... "

Package.swift
```diff
@@ -0,0 +1,42 @@
+// swift-tools-version:5.0
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "CocoaLumberjack",
+    products: [
```
ffried: We should add our platform requirements.

Classes/DDASLLogCapture.m
```diff
@@ -13,7 +13,7 @@
 //   to endorse or promote products derived from this software without specific
 //   prior written permission of Deusty, LLC.
 
-#import "DDASLLogCapture.h"
+#import <CocoaLumberjack/DDASLLogCapture.h>
```
ffried: Excerpt from the SE proposal that introduced C-language support:

>When building a target, the package manager will automatically add an additional header search path argument to the include directory for each C language target in the transitive closure of the dependencies for the target being built.
>
>We should use `-iquote` as the header search argument for targets which are within the current package, and `-I` for targets which come from package dependencies. This allows projects to use `#include "foo/foo.h"` versus `#include <foo/foo.h>` syntax appropriately to distinguish between the inclusion of headers which are or are not within the package.

So I think there should be a way to continue including our headers using normal quotes instead of angle brackets.

.gitignore
```diff
@@ -1,23 +1,12 @@
...
-xcuserdata
-*.xccheckout
-*.moved-aside
-DerivedData
-*.hmap
-*.ipa
-*.xcuserstate
+xcuserdata/
+
+**/*.xcodeproj/project.xcworkspace/*
```
ffried: I'm not sure that it's a good idea to exclude everything of a workspace except the shared data. The internal workspace structure is arbitrary and Apple may change it at any time. It wouldn't be the first time they add new files that are then added for everyone using the project - potentially causing a confirmation dialog in their projects.
We should simply ignore the `xcuserdata`, but include all the other things.
For SPM support, we should also explicitly exclude the project generated by SPM and move our Xcode project into a subfolder (e.g. Xcode).

.gitignore
```diff
@@ -32,3 +21,7 @@ Pods/
 Podfile.lock
 LumberjackUser*.h
 .DS_Store
+
+# SwiftPM
+.swiftpm
+.build
```
ffried: NitPickk: Missing newline at end of file

Classes/DDASLLogCapture.m
diederich: Quote includes in .m files should be fine I think. On the other hand, if they all use the same style it shouldn't matter.

diederich:
> @diederich Where do you put `CocoaLumberjackSwift.h`?
> Its include statements seems to be broken. :/

I created  `Classes/include/CocoaLumberjack` and put all the public headers into it,  including `CocoaLumberjack/CocoaLumberjack.h`.

lolgear:
@diederich 
What about "CocoaLumberjackSwift.h"? 
I still can't figure out where it should be.

.gitignore
lolgear: @ffried Is there any option for DangerFile which would check newlines at end of file?

diederich:
> @diederich
> What about "CocoaLumberjackSwift.h"?
> I still can't figure out where it should be.

What @sushichop said worked for me as well.
Putting all public headers, including CocoaLumberjack.h into `Classes/include/CocoaLumberjack`.
Then, in Package.swift, specify `publicHeadersPath : "Classes/include"`. 
Does that work for you?

lolgear:
@diederich 
I don't have troubles with `CocoaLumberjack.h`. 
The main problem is `CocoaLumberjackSwift.h`. It should be, I suppose, under `Classes/Include/CocoaLumberjackSwift` if option for separate swift target is necessary via swift-pm.

.gitignore
lolgear: @ffried I am asking, because I know that you would like to separate it into another PR :)

diederich:
> @diederich
> I don't have troubles with `CocoaLumberjack.h`.

Oh man, sorry. I've been misreading this...

> The main problem is `CocoaLumberjackSwift.h`. It should be, I suppose, under `Classes/Include/CocoaLumberjackSwift` if option for separate swift target is necessary via swift-pm.

The swift module itself doesn't need a header file. Public/Private is done via attributes. The only include in that header file seems to be the log-level defaults? I'm not even sure anymore if we need a separate swift target TBH..

.gitignore
lolgear: @ffried I copied this content from WWDC 2019 ( [`BuildingCustomViewsInSwiftUI`](https://developer.apple.com/documentation/swiftui/drawing_and_animation/building_custom_views_in_swiftui) ).

Package.swift
lolgear: @ffried 
AppleOS requirements?

ffried: Yep, see [here](https://github.com/ffried/FFFoundation/blob/a645fe75168926fed06fe8bd67cfebb388603919/Package.swift#L8-L13) for an example.

lolgear: @ffried 
I have strong desire to add `Package-SupportPlatforms.swift` OR `SupportedPlatforms.yaml` and read them in `Package.swift` and in `.podspec`.

.gitignore
ffried: I don't know if it's clever to use a UI talk for `.gitignore` files 😉 
If you find some documentation from Apple recommending this, I'm fine with it. I'm just talking out of experience, that it's extremely cumbersome if e.g. the file with the 32-bit warning is ignored and devs get a confirmation dialog every time they open the project.

.gitignore
ffried: I think Danger could do this, yea. And yea, you're right, this is scope of another PR. 🙂 

Since you get the changed files, you could probably walk through them and check if they have a newline at the end of the document.

.gitignore
lolgear: @ffried 
I know one correct way to get ".gitignore" that Apple recommends - create repository for a new project. But.. I don't use Xcode for Git tasks :)

Package.swift
ffried: IMHO this is too much hassle for something that rarely changes.

.gitignore
ffried: Yea, but the `.gitignore` SPM generates is a lot simpler.

.gitignore
lolgear: @ffried 
Or I can setup local DangerFile with newline checks and ask it to do that. ;)

Classes/DDASLLogCapture.m
lolgear: @diederich @ffried 
"Include of non-modular header inside framework ..." everywhere.
Yes, `#import "CocoaLumberjack/DDLog.h"` works fine with Swift Package, but it breaks Xcode build.

lolgear:
@CocoaLumberjack/collaborators 
Any ideas how to solve angle-brackets/quotes imports puzzle?

Here it is.
Umbrella header should have import statements with angle-brackets. Each public header that is mentioned in umbrella header also should have only angle-brackets imports.

However, angle-brackets is an old way to import headers. Xcode uses smart search paths to #import <Library/Library.h> and #import "Library.h" and autocomplete relative paths.

ffried:
@lolgear So, I've created a small test project-setup that more or less mirrors our situation.
Almost everything is working fine. The only thing I can't get to work is the ObjC tests in SPM. `swift build` works fine, but `swift test` doesn't find the imported headers.
The Xcode-Project (made for "normal" app targets) works great however. Tests run fine and so do builds.

I've attached the project here, in case you want to play around with it as well:
[ObjCPackage.zip](https://github.com/CocoaLumberjack/CocoaLumberjack/files/3397234/ObjCPackage.zip)

A few notes on that:
- I've re-structured the file layout to follow SPMs setup. This makes it easier to maintain the SPM support and I do like the file structure SPM uses (which is a personal opinion and might not match what we'll decide here)
- All "internal" imports have the "include" folder inside it (`#import "include/Header.h"`). While I agree that it might not be what "modern ObjC" looks like, it's what C/C++ still uses and therefor what is needed for SPM.
- All "external" imports (cross module) are in the normal `#import <Module/Header.h>` style.

My sample project also contains two targets: one for ObjC and one for Swift. The latter depends on the first - just like in our case.

ffried:
Oh, a quick note on the "additions" to the SPM file structure that I've made:

- There's a folder called `Xcode` on the top level, that contains the Xcode project for "normal" app targets. This is to prevent collisions with the Xcode Project generated by SPM.
- All the files that are only needed by the "normal" app targets are inside a folder called "Supporting Files" (following Xcode's naming) inside the folder for each target. This folder is then excluded inside the Package.swift, so that SPM does not further process them.

lolgear:
@ffried 
Yes, `swift test` doesn't work even after changing to `@import ObjcPackage;`.  :(

Do you have idea how to sort headers according to their access level? You have only public headers here.

ffried:
@lolgear Yea, I've tried this, too. I'm not sure what SPM needs to lookup the dependencies for an ObjC test target. Might even be a bug in SPM.

project / private headers are irrelevant. You can put them wherever you want. I'd put them into the same folder where they are needed, allowing source files to import them using `#import "Header.h"`.

lolgear:
@ffried 
I suggest to put them into `include_private` folder, for example. Or even `PrivateInclude` to mimic `PrivateHeaders` structure from Xcode derived data.

ffried:
@lolgear I'm actually not sure about that. SPM's support for ObjC is based on its support for C/C++. So IMHO we should structure our project the way C/C++ projects do. Taking the Swift compiler as an example, they also store the private headers in the same folder as the source files using them.

lolgear:
@ffried 
You have disabled "Quoted include in framework header" warning.
I have tested public includes of headers into each other. It doesn't work.

1. Add MyDerivedObject.h ( Public access level )
2. Add `#import <ObjCObject/MyBaseObject.h>`
3. Compile project in Package.swift
4. Get error :(

ffried:
@lolgear Yea, because it's not an "external" import - as in it's not cross-module. Therefore you need to use the quoted version: `#import "MyBaseObject.h"`
Since you're already inside the "include" folder, you also don't to specify the path to the header.

> You have disabled "Quoted include in framework header" warning.

I've just enabled it and it still builds without warnings. This is because the "framework header" is the "umbrella header". In the ObjCPackage case, this is the header file inside the Supporting Files folder. There, you need to use angle brackets for all imports - adding `#import <ObjCPackage/MyDerivedObject.h>` there works fine btw.

Now, I'm not sure if the quoted imports in the public header files is causing problems with e.g. CocoaPods. If so, we probably need to use a preprocessor macro for the imports, together with a define that tells us if we're in a build mode that needs angle bracket imports there. But I'd first try to use the same style for all integration methods.

ffried:
*Update on the warning*:
Seems like my Xcode had a bad start this morning. Was somehow stuck in using a build cache. Now the warning shows up, yea...

ffried:
But the fix is easy. Just apply the fix-it Xcode provides and change `#import "MyBaseObject.h"` to `#import <MyBaseObject.h>`. 🙂 
Builds fine with SPM and Xcode.

ffried:
So, I've come up with a possible "solution" to the import problem.
I've added some macros and build settings, that dynamically form the correct import statement. There's a catch, however. The macros reside in their own header file - which needs to be imported. So For this particular import, we're stuck with `#if - #else - #endif`-wrapping it.

The downside of all of it is, that imports are no longer as straightforward as they used to be. The upside however is, that in the preprocessed source files, the import are always correct and don't have to follow e.g. the C-style for an Xcode target (which could use normal ObjC style).

Attached the current version of the sample project:
[ObjCPackage.zip](https://github.com/CocoaLumberjack/CocoaLumberjack/files/3400512/ObjCPackage.zip)

lolgear:
@ffried 
Good morning :)
I have suggested similar solution above, but @diederich responded not to hack headers imports. Reasonable enough.

At the end we have only one solution - inconsistent headers handling, right?

ffried:
@lolgear @diederich I agree that it's not optimal. On the other hand, we should either fully support SPM or not at all IMHO. Even if we found a way to leave the imports as is, it might break at any time.

lolgear:
@ffried 
But we have another option...
We can class/protocol/symbol-forwarding everything in public headers across all headers...
For example, `DDFileLogger.h` has `#import <CocoaLumberjack/DDLog.h>`. It is ridiculous. File logger doesn't depend on logging public API...

ffried:
@lolgear Nice idea, but won't work for superclasses.

```objective-c
@class MyBaseClass;

@interface MySubclass: MyBaseClass
@end 
```
This fails, because superclass constraints can't be made with forward declarations.

Also, I don't like the idea of "disabling" a feature (imports) just because they don't work with a particular integration method.

lolgear:
@ffried 

```
// .h
@interface DDFileLoggerBuilder : NSObject
+ (DDFileLogger *)createLogger;
@end

// .m
@intreface DDFileLogger : DDBaseLogger
@end
```

ffried:
@lolgear Introducing wrapper classes (increasing the binary interface of the framework and maintenance "cost") just to make SPM work? IMHO this is far worse than hacking our imports a little... 😉

lolgear:
@ffried 
Not sure that wrapper classes are worse than subclassing. Even buffering for file logger has been made by proxy.
`File[Logging]` vs `FileLoggerSubclass`.
But denying imports is the worst solution :)

ffried:
Wrapper classes that make things simpler are fine. But in this case they only serve the purpose of enabling SPM. They're not needed otherwise. So in this case, they're unnecessary.

I'll play around a little more. Maybe I'll find an acceptable solution.

lolgear:
@ffried 
I have submitted a bug to SPM several days ago... I would like to know how to setup ObjC libraries with all possible options. No responses. :(

diederich:
The last issue was not being able to build the tests via `swift test`?

lolgear:
@diederich 
The last issue is import statements are totally wrong for SPM :)
Do you have some clues about `swift test`?

ffried:
> I have submitted a bug to SPM several days ago... I would like to know how to setup ObjC libraries with all possible options. No responses. :(

@lolgear Can you post a link to said bug? Maybe the forums are a better place to ask for this.

@diederich @lolgear I've solved the test problem. Also import related. They need to be normal quote imports there. `swift test -v` is a big help, showing what underlying commands it actually generates.

I'm also a little closer on solving the import problem in normal source files. Similar to what @diederich did in the screenshot above, but with slightly different build settings.

ffried:
@lolgear But I think ObjC Targets in SPM don't get much priority at Apple. SPM is not really made for ObjC. It's just able to somewhat support it, since ObjC is mostly using the same compiler as C/C++ on Darwin platforms.

lolgear:
@ffried 
[Add comprehensive examples for existed Open Source Objective-C projects](https://bugs.swift.org/browse/SR-11076?jql=project%20%3D%20SR%20AND%20status%20%3D%20Open%20AND%20component%20%3D%20%22Package%20Manager%22)

Yes, ObjC is a zombie since 2014-2016. At this year it's finally dead by introduction of Combine/SwiftUI.

sushichop:
> @lolgear But I think ObjC Targets in SPM don't get much priority at Apple. SPM is not really made for ObjC. It's just able to somewhat support it, since ObjC is mostly using the same compiler as C/C++ on Darwin platforms.

I absolutely agree with @ffried . And IMHO, I think we are only able to import not angle brackets header but simple header at test target using `swift test` **for now**.

e.g.(test target about above `ObjCPackage.zip`)
Use import "MyBaseObject.h" instead of  `#import <ObjCPackage/MyBaseObject.h>`.
Use import "MyDerivedObject.h" `#import <ObjCPackage/MyDerivedObject.h>`.

sushichop:
I'm afraid of CocoaLumberjackSwift target. Probably I think we must add `import CocoaLumaberjack` for each swift files.

ffried:
> I'm afraid of CocoaLumberjackSwift target. Probably I think we must add import CocoaLumaberjack for each swift files.

@sushichop Which swift files? Ours? That wouldn't be a big deal. There aren't many of them 😉

sushichop:
@ffried 
You are right. There was no need to worry too much about number swift files😅 Just two files.
https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/Classes/CocoaLumberjack.swift
https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/Classes/DDAssert.swift

But we may need to be careful about take care of how to importing... using CocoaPods integration...
(CocoaLumberjack/Swift sub spec.)

ffried:
@sushichop Since there are only two files, we could wrap the imports into compiler if's there:
```swift
#if SWIFT_PACKAGE
import CocoaLumberjack
#endif
```
But only if it really becomes a problem of course. Right now my bigger concern is getting the ObjC part to work without too much hassle.

sushichop:
Maybe we need to add method to get `DDDefaultLogLeve`l value in SwiftLogLevel.h for  CocoaLumberjackSwift target. Because Swift supports only simple MACRO.
Though I haven't check yet...(sorry I don't have enough to spare time recently😅)

ffried:
@sushichop This is what the `static const DDLogLevel DDDefaultLogLevel = DD_LOG_LEVEL;` is for. Swift itself only sees constants, but the header files can still contain defines. The trick here is to turn the define into a constant - which is also evaluated as constant by the compiler, allowing strings to be stripped if they're not logged. But this is nothing new and should work with SPM just like it does with the current Xcode targets.

ffried:
@lolgear @diederich @sushichop Good news. I finally have a project that works fine!

[ObjCPackage.zip](https://github.com/CocoaLumberjack/CocoaLumberjack/files/3401782/ObjCPackage.zip)

No additional linker settings are necessary - neither in the Xcode targets nor in SPM. Both import statements are supported `#import <Library/Header.h>` for module headers, and `#import "Header.h"` for implementation files.
The same goes for tests. `swift test` works fine. Imports in tests can be either `#import <Library/Header.h>` or `#import "Header.h"` - both work.

The trick is to create a symlink with the same name as the module inside the `include` folder. 🙂

sushichop:
> Good news. I finally have a project that works fine!

Great!

sushichop:
@ffried 
> This is what the static const DDLogLevel DDDefaultLogLevel = DD_LOG_LEVEL; is for. Swift itself only sees constants, but the header files can still contain defines. The trick here is to turn the define into a constant - which is also evaluated as constant by the compiler, allowing strings to be stripped if they're not logged. But this is nothing new and should work with SPM just like it does with the current Xcode targets.

Thanks! I checked it works fine in Xcode before. But I'm still not sure it works using `swift build`. Because I suppose `CocoaLubmerjajckSwift` target depends on `CocoaLumberjack` target(only Objective-C) when we use SwiftPM. IMHO, for example, I think we maybe replace  DDDefaultLogLevel to CocoaLubmeracjack.DDDefaultLogLevel? 
Anyway I'll check it later when I have time🙂

ffried:
> Thanks! I checked it works fine in Xcode before. But I'm still not sure it works using swift build. Because I suppose CocoaLubmerjajckSwift target depends on CocoaLumberjack target(only Objective-C) when we use SwiftPM. IMHO, for example, I think we maybe replace DDDefaultLogLevel to CocoaLubmeracjack.DDDefaultLogLevel?

@sushichop The dependency layout is the same. Today, the swift target also depends on the ObjC target. The same goes for SPM. The `SwiftLogLevel.h` header is part of the Swift target, not the ObjC target. Thus it can use the define from the (imported) ObjC target and convert it into a constant that can then be used by Swift. It's basically the bridge between the ObjC define and the swift constant.

sushichop:
@ffried  OK! It's perfect😄

diederich:
I'm probably missing something, but I've replaced the symlink with an actual folder (named like the module), put the public headers as well as the umbrella header into it and used angle-braket includes for a .m file. 
I built the targets in the Xcode project, and ran swift test. Anything else to test?
[ObjCPackage.zip](https://github.com/CocoaLumberjack/CocoaLumberjack/files/3401904/ObjCPackage.zip)

ffried:
@diederich The current setup of imports might work with that. But there are multiple downsides to that:
- It makes the file structure more complicated and repetitive. If possible, I'd like the Xcode structure to match the file system. But then we have an additional folder with the same name as the target within the include folder. Not nice IMHO.
- It disables support of quote imports completely (which is how you usually import headers in implementation files).

The symlink is hidden inside the Xcode structure. Therefor it's quite a simple move to enable the complete flexibility.

Btw. we should use an umbrella header only for the Xcode targets. SPM generates it's own umbrella header.

diederich:
> * It makes the file structure more complicated and repetitive. If possible, I'd like the Xcode structure to match the file system. But then we have an additional folder with the same name as the target within the include folder. Not nice IMHO.

It's one of the solutions proposed in SwiftPM, that's where the idea came from:

> If include/Foo/Foo.h exists and Foo is the only directory under the include directory then include/Foo/Foo.h becomes the umbrella header.

https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md

I don't have a strong opinion on the symlink solution vs. include/ModuleName. Though the recursiveness of the symlink makes me feel a bit uncomfortable :-) 

> * It disables support of quote imports completely (which is how you usually import headers in implementation files).

Well, you could do relative paths like `#import "include/...`, but probably nicer would be to setup a header search path pointing into the include/ModuleName folder?

[snip]

> Btw. we should use an umbrella header only for the Xcode targets. SPM generates it's own umbrella header.

We could also reuse the one we have for both variants to make sure to provide the same interface?

ffried:
> It's one of the solutions proposed in SwiftPM, that's where the idea came from:
>
>> If include/Foo/Foo.h exists and Foo is the only directory under the include directory then include/Foo/Foo.h becomes the umbrella header.
https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md
>
> I don't have a strong opinion on the symlink solution vs. include/ModuleName. Though the recursiveness of the symlink makes me feel a bit uncomfortable :-)

I know. But with symlinks we use the same rule - just without actually moving the header files.
The recursiveness is a little downside, yes. But since the compiler collects absolute paths, it should work fine. I'm not even sure if the compiler walks down more than one level there.

> Well, you could do relative paths like #import "include/..., but probably nicer would be to setup a header search path pointing into the include/ModuleName folder?

I've had the header search path version working. But again, I'd like to stay as close to the plain setup as possible as it minimizes A) the initial effort to support SPM and B) the continuous effort needed to keep up that support.

> We could also reuse the one we have for both variants to make sure to provide the same interface?

It would work, yes. But there are few downsides to that, too. One being playground support, which apparently only works if SPM generates the umbrella header and module maps.
I still wonder why they're still needed for Xcode targets. If SPM is able to generate them, why isn't Xcode? 🧐

lolgear:
@ffried 
Well, good news - ObjC target has been fixed by this nice-to-know hack :)
Bad news - I still can't fix Swift target ( It doesn't know about DefaultLogLevel ). Probably, it doesn't know how to find its headers.

ffried:
@lolgear Did you put the `SwiftLogLevel.h` file in to the Swift target? We'll probably need to provide the umbrella header here, since it needs to import this header.

lolgear:
@ffried I do.
I put `CocoaLumberjackSwift.h` and `SwiftLogLevel.h` in `CocoaLumberjackSwift/include/CocoaLumberjackSwift`.

diederich:
> It would work, yes. But there are few downsides to that, too. One being playground support, which apparently only works if SPM generates the umbrella header and module maps.

Oh, that's a good call. 🙏

ffried:
> @ffried I do.
> I put CocoaLumberjackSwift.h and SwiftLogLevel.h in CocoaLumberjackSwift/include/CocoaLumberjackSwift.

My bad. Extra header files don't work in Swift targets in SPM. What we need is an intermediate target for the header files. See https://github.com/lolgear/CocoaLumberjack/pull/1 😉

.gitignore
```diff
@@ -1,34 +1,17 @@
...
-#
-Pods/
-Podfile.lock
+# Lumberjack generated files
 LumberjackUser*.h
-.DS_Store
+
+# SwiftPM
+.swiftpm
+.build
```
ffried: `/.build` is already excluded above.
We should also only exclude `.swiftpm` on the root level (as in `/.swiftpm`) and move it up to the `/.build` definition.

CocoaLumberjack.podspec
```diff
@@ -28,13 +28,13 @@ Pod::Spec.new do |s|
   s.default_subspecs = 'Core'
 
   s.subspec 'Core' do |ss|
-    ss.source_files         = 'Classes/CocoaLumberjack.h', 'Classes/DD*.{h,m}', 'Classes/Extensions/*.{h,m}', 'Classes/CLI/*.{h,m}'
-    ss.private_header_files = 'Classes/DD*Internal.{h}'
+    ss.source_files         = 'Classes/Include/**/*.{h}', 'Classes/DD*.{h,m}', 'Classes/Extensions/*.{h,m}', 'Classes/CLI/*.{h,m}'
```
ffried: the Podspec no longer matches the folder structure

lolgear: Yes...

ffried:
Apart from the two things above, it looks already pretty solid now. The Tests are still left. With the symlink version in place, they should be able to build fine, too, as long as we're using `#import <CocoaLumberjack/Header.h>` or `#import "Header.h"`...

lolgear:
@ffried 
No, they are not...
I run command `swift test`.
It fails "Module not found" in both `@import CocoaLumberjack;` and `#import <CocoaLumberjack/CocoaLumberjack.h>`

ffried:
@lolgear Yes, you cannot import the (generated) umbrella header. But tests should also usually not do so. Instead they should import the headers of the objects they test.
So you'd had to e.g. `#import <CocoaLumberjack/DDFileLogger.h>` or `#import "DDFileLogger.h"`.

Lumberjack.xcodeproj/project.pbxproj
```diff
@@ -95,77 +90,77 @@
...
-		0A3F9CE921E3F8AF00F3B707 /* DDFileLogger+Internal.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "DDFileLogger+Internal.h"; sourceTree = "<group>"; };
-		0AA59E1721DD2A080031787C /* DDFileLogger+Buffering.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "DDFileLogger+Buffering.h"; sourceTree = "<group>"; };
-		0AA59E1821DD2A080031787C /* DDFileLogger+Buffering.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = "DDFileLogger+Buffering.m"; sourceTree = "<group>"; };
-		0AE6D52321941F0900B2A35D /* DDLoggerNames.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = DDLoggerNames.m; sourceTree = "<group>"; };
-		0AE6D52421941F0900B2A35D /* DDLoggerNames.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = DDLoggerNames.h; sourceTree = "<group>"; };
+		0752523B22E075E1005D30B3 /* SwiftLogLevel.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = SwiftLogLevel.h; path = Sources/CocoaLumberjackSwiftSupport/include/SwiftLogLevel.h; sourceTree = SOURCE_ROOT; };
```
OSSPolice: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/CocoaLumberjack/CocoaLumberjack/blob/3d238d3b0cfb23ee822aca5b89edfae867aa19e4/Lumberjack.xcodeproj/project.pbxproj#L118"></span>Files should be in sync with project structure</td>
    </tr>
  </tbody>
</table>

ffried: I agree. But unfortunately we need this exception here... Sorry my friend... ¯\_(ツ)_/¯

Lumberjack.xcodeproj/project.pbxproj
```diff
@@ -95,77 +95,77 @@
...
```
OSSPolice: <table data-meta="generated_by_danger">
  <tbody>
    <tr>
      <td>:warning:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/CocoaLumberjack/CocoaLumberjack/blob/40371aaae76c6a0e7194a8b50db3f6cd0718d85a/Lumberjack.xcodeproj/project.pbxproj#L123"></span>Files should be in sync with project structure</td>
    </tr>
  </tbody>
</table>

lolgear:
@ffried 
Well... CocoaPods won't build :(

```
    - ERROR | [CocoaLumberjack/Swift] xcodebuild:  /Users/dmitry/Programming/GitHub/ObjectiveC/CocoaLumberjack/Sources/CocoaLumberjackSwift/Supporting Files/CocoaLumberjackSwift.h:19:2: error: import of module 'CocoaLumberjack' appears within same top-level module 'CocoaLumberjack'
    - ERROR | [CocoaLumberjack/Swift] xcodebuild:  /Users/dmitry/Programming/GitHub/ObjectiveC/CocoaLumberjack/Sources/CocoaLumberjackSwift/Supporting Files/CocoaLumberjackSwift.h:21:9: error: 'CocoaLumberjackSwift/SwiftLogLevel.h' file not found
```

I suppose it requires another subspec to match `SwiftLogLevel.h` new location...

ffried:
@lolgear Nah, you're just including the umbrella header now, which wasn't done before.
The following subspec defs work:

```ruby
  s.subspec 'Core' do |ss|
    ss.source_files         = 'Sources/CocoaLumberjack/Supporting Files/*.h', 'Sources/CocoaLumberjack/include/*.h', 'Sources/CocoaLumberjack/*.{h,m}'
    ss.private_header_files = 'Sources/CocoaLumberjack/DD*Internal.{h}'
  end

  s.subspec 'Swift' do |ss|
    ss.dependency 'CocoaLumberjack/Core'
    ss.source_files = 'Sources/CocoaLumberjackSwiftSupport/include/*.h', 'Sources/CocoaLumberjackSwift/*.swift'
  end
```

ffried:
@lolgear Also, you need to move the `import CocoaLumberjack` in the two Swift files into the `#if SWIFT_PACKAGE` block. Otherwise the CocoaPod build emits two warnings, causing `pod lib lint` to fail.

CocoaLumberjack.podspec
```diff
@@ -28,13 +27,14 @@ Pod::Spec.new do |s|
   s.default_subspecs = 'Core'
 
   s.subspec 'Core' do |ss|
-    ss.source_files         = 'Classes/CocoaLumberjack.h', 'Classes/DD*.{h,m}', 'Classes/Extensions/*.{h,m}', 'Classes/CLI/*.{h,m}'
-    ss.private_header_files = 'Classes/DD*Internal.{h}'
+    ss.source_files         = 'Sources/CocoaLumberjack/**/*.{h,m}'
+    ss.private_header_files = 'Sources/CocoaLumberjack/DD*Internal.{h}'
+    ss.public_header_files  = 'Sources/CocoaLumberjack/include/*.{h}', 'Sources/CocoaLumberjack/Supporting Files/*.{h}'
```
ffried: Explicitly defining public header files isn't necessary. CocoaPods treats all headers as public unless they're present in `private_header_files`

Package.swift
```diff
@@ -0,0 +1,39 @@
...
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
+        .target(name: "CocoaLumberjack",
+                exclude: ["Supporting Files"]),
+        .target(name: "CocoaLumberjackSwiftSupport",
+                dependencies: ["CocoaLumberjack"]),
+        .target(name: "CocoaLumberjackSwift",
+                dependencies: ["CocoaLumberjack", "CocoaLumberjackSwiftSupport"],
+                exclude: ["Supporting Files"]),
+        .testTarget(name: "Tests",
```
ffried: We should name the tests `"CocoaLumberjackTests"` since they currently only test the ObjC target. This makes it easier to add tests for the Swift target in the future.

Package.swift
```diff
@@ -0,0 +1,39 @@
...
+                    dependencies: ["CocoaLumberjack"],
+                    path: "Tests",
```
ffried: Do we need `path`, `exclude`, and `sources`? If we name the folders correctly, SPM should catch this automatically.

lolgear: @ffried 
For that we need to `move Tests -> Sources/CocoaLumberjackTests`, right?

ffried: Nope. Tests are always searched inside the "Tests" folder. So you need to have a folder "CocoaLumberjackTests" inside "Tests" and it'll automatically pick those up.

lolgear: @ffried 
I don't remember if it is required or not to include `.testTarget()` in `.library()` in SPM.

ffried: @lolgear You mean if you need to define a product for it? You don't. 
Targets are internal and don't leave the package. Products are what leave the package. But tests should never leave the package. They should only be internal to the package.

lolgear:
@ffried @CocoaLumberjack/collaborators 
Well, it is done.
The main question is "Should we merge it"?

Sources/CocoaLumberjack/include/DDTTYLogger.h
```diff
@@ -31,7 +31,7 @@
     static inline DDColor* DDMakeColor(CGFloat r, CGFloat g, CGFloat b) {return [DDColor colorWithRed:(r/255.0f) green:(g/255.0f) blue:(b/255.0f) alpha:1.0f];}
 #elif defined(DD_CLI) || !__has_include(<AppKit/NSColor.h>)
     // OS X CLI
-    #import <CocoaLumberjack/CLIColor.h>
+    #import "CLIColor.h"
```
ffried: Was this import change on purpose? It's not relevant for SPM and is inside a public header. So it should use angle bracket style, right?

ffried:
@lolgear What do you mean "merge"? This was a nice experiment, now let's just throw it away... 😛 

jk... IMHO this looks great now! I'm just not sure if this means a major version jump. While we have almost no breaking changes, moving the Xcode project from `Framework` to `Xcode` is one. We could provide a symlink from `Framework` to `Xcode` to work around this for now.

I'll see if I can test that in one of our production apps next week. If nothing else breaks, I'd say this is a very big step forward!

In either way: Thanks for putting this together @lolgear!

Sources/CocoaLumberjack/include/DDTTYLogger.h
lolgear: @ffried 
It was one of the steps in experiment that I forget to reverse :)

lolgear:
@ffried 
I would like to put .xcproj to Xcode/ also. And...
I am not sure that `Framework/` directory has greater discoverability than `Framework/`.

ffried:
@lolgear It's less about the discoverability (although I find `Xcode` is better discoverable than `Framework`, because I'd expect binaries in the latter), more about breaking existing projects.

ffried:
@lolgear I'd suggest to put the .xcodeproj also in `Xcode` and then simply create a symlink from `/Lumberjack.xcodeproj` to `/Xcode/Lumberjack.xcodeproj` and mention the deprecation of the `/Lumberjack.xcodeproj` location in the release notes. I think, with that change, the addition of SwiftPM support is completely non-breaking.

ffried:
@lolgear I've also just verified this with one of the bigger projects I'm using CocoaLumberjack in. Switched the dependency to your branch and it still builds just fine!

sushichop:
Great works!
I also confirmed `swift test` works fine(including `swift build`). And CocoaPods and Carthage integration works fine as before🙂

But... I could not confirm the following steps work fine(I could not build using Xcode and I needed to add AppKit.framework for CocoaLumberjack.framework target to build it using Xcode...😢). 
Is this a limitation or do I need Xcode 11 for SwiftPM to work completely?
My Xcode version is 10.3. 

```
swift package init 
swift build
swift package generate-xcodeproj
open Dependencies.xcodeproj
(Try to build using Xcode.)
```

Package.swift
```swift
// swift-tools-version:5.0
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "Dependencies",
    products: [
        // Products define the executables and libraries produced by a package, and make them visible to other packages.
        .library(
            name: "Dependencies",
            targets: ["Dependencies"]),
    ],
    dependencies: [
        // Dependencies declare other packages that this package depends on.
        .package(url: "git@github.com:lolgear/CocoaLumberjack.git", .branch("swift_package")),
    ],
    targets: [
        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
        .target(
            name: "Dependencies",
            dependencies: ["CocoaLumberjackSwift"]),
        .testTarget(
            name: "DependenciesTests",
            dependencies: ["Dependencies"]),
    ]
)
```

And I drag-and-dropped the above Dependencies.xcodeproj to some macOS app project. 
Then I could import CocoaLumberjack.framework, but I could not import CocoaLumberjackSwift.framework...😰 

As a whole, am I wrong with how to use SwiftPM?

ffried:
@sushichop Wow, thanks for testing this. I totally missed that one. I only tested building CocoaLumberjack directly... 😨 

Just to clarify: Building your own package on the terminal (using `swift build`) works fine, but the generated Xcode project doesn't build? If so, that's most likely a bug in SwiftPM. `swift build` and `xcodebuild -project {GENERATED_XCODEPROJ} build` should yield the same result.

We could work around it by conditionally linking AppKit (conditional, since we cannot link it on non-macOS targets).

Can you test Xcode 11? If it's fixed there, I'd simply declare SPM support as of Xcode 11.

I'd say drag-and-dropping the project is not really a valid SPM case. You'd be creating another package and declaring a dependency there.

dehlen:
I tested on Xcode 11 beta 4 by adding https://github.com/lolgear/CocoaLumberjack.git, branch: swift_package as a new Swift Package dependency. The imports work fine for me.

ffried:
@dehlen So the generated Xcode project builds fine, too?

dehlen:
When I right click the package -> Show in Finder -> and open the Lumberjack.xcodeproj in Xcode 11 beta 4 the CocoaLumberjack and CocoaLumberjackSwift Target build succeeds.

bpoplauschi:
You guys had a very good conversation here, including testing.
Any blockers for merging the PR?

bpoplauschi:
You guys had a very good conversation here, including testing.
Any blockers for merging the PR?

lolgear:
@bpoplauschi 
I think we are ready, but we should add new Integration projects, I suppose?
Here we have exactly the same situation with CocoaPods.
`pod lib lint` feels good, so, we don't integrate anything to test deployment in CocoaPods environment.
`swift test` feels good... and we should do something else?
Maybe `Demos` projects need another entry.

ffried:
@lolgear I'd say for the initial support we don't need any new samples. SPM is pretty straightforward. I'd only include it in the README for now and add samples as necessary.

Apart from that I'd really like to move the .xcodeproj into the Xcode folder and (for now) provide a symlink) to prevent confusion with the SPM generated xcodeproj - which should also be added in the .gitignore as `/*.xcodeproj`.

sushichop:
Hi guys!
I confirmed SwiftPM works fine using Xcode 11 Beta 5!🙂

I created Package.swift from Xcode menu.
(However I want to avoid taking screenshots because, IMHO, it may violate Apple license agreement😎)

Package.swift is here.
```
// swift-tools-version:5.1
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "MyLibrary",
    platforms: [.iOS(.v8)],
    products: [
        .library(
            name: "MyLibrary",
            targets: ["MyLibrary"]),
    ],
    dependencies: [
        .package(url: "https://github.com/lolgear/CocoaLumberjack.git", .branch("swift_package"))
    ],
    targets: [
        .target(
            name: "MyLibrary",
            dependencies: ["CocoaLumberjackSwift"]),
        .testTarget(
            name: "MyLibraryTests",
            dependencies: ["MyLibrary"]),
    ]
)
```

On the other hand `swift package generate-xcodeproj` did not work fine. 
It may be because it is still beta version... I'm not sure, but I think this is NOT a problem🙂

So I'm glad that we can say CocoaLumberjack will completely support SwiftPM with Xcode 11(Sep. this year?)🎉

Can I add README to this PR or create another PR for it?
It may be enough to add the description of Pakcage.swift and Xcode version to README😅

lolgear:
@suchichop
I suggest to add full setup with even “In Xcode choose menu item File -> choose add repository -> insert github link to this repository” and etc.
For me it was a big surprise that I didn’t know how to start with SPM in Xcode 11 beta 1.

About PR..
I also suggest to PR into this PR.

Also, if you like to complete final steps... Do you dare to move Xcode project to subdirectory? I don’t know if Apple have plans to release migration tool which allows to move Xcode project and repair relative paths of files in it.

ffried:
@sushichop @lolgear IMHO it's not our job to provide an intro on "how to use SPM with Xcode". That's Apple's job. If you don't know how to use SPM, first learn that, then come back and include CocoaLumberjack as dependency.

Most SPM projects (including Apple's own Swift NIO) simply list their dependency-line and I think this is fine. E.g. we also don't provide instructions on how to install Carthage oder how to setup CocoaPods:
- [Vapor](https://docs.vapor.codes/3.0/getting-started/spm/)
- [Swift NIO](https://github.com/apple/swift-nio#getting-started)
- [YAMS](https://github.com/jpsim/Yams#swift-package-manager)
- [XMLWrangler](https://github.com/sersoft-gmbh/XMLWrangler#installation) (DISCLAIMER: I'm maintaining that project!)
- [Alamofire](https://github.com/Alamofire/Alamofire#swift-package-manager)
- [SwiftDate](https://github.com/malcommac/SwiftDate/blob/master/Documentation/0.Informations.md#swift-package-manager)
- [PromiseKit](https://github.com/mxcl/PromiseKit/blob/master/Documentation/Installation.md#accio)
- ...


Regarding the move of the Xcode project: Since our package is named "CocoaLumberjack", SPM will generate a "CocoaLumberjack.xcodeproj". The Xcode project we currently have is named "Lumberjack.xcodeproj". This means that for now it would work, but I assume it could cause irritation on what is what... Given all the custom settings we have with the xcconfig files, I'd rather not rely on the SPM generated project.
But maybe that's a topic for a different PR altogether. So let's just put this aside for now.

sushichop:
@ffried @lolgear 
Thanks for sharing your idea. I'm agree with @ffried . 
Then can you add CHANGELOG before we merge this PR to master branch?

ffried:
@amayers Do the other packages have internal dependencies? `CocoaLumberjackSwiftSupport` is an internal **target** that is not exposed via a **product**. I assume this is what is still causing issues with Xcode.

ffried:
@amayers Thanks for the confirmation. I'll try to create a small reproduction project that is not using any external dependencies and file a radar with Apple about that.

ffried:
I'll update the Changelog on master directly.

PR #1098, opened 2019-09-25
fixed an issue with log archiving in the simulator

mackuba:
This fixes a regression introduced in d0fed54ea39d710918f5b9d3627d32a50f889c46. The code that checks the previously used NSFileProtectionType key needs to check if the returned key isn't nil, which happens in the simulator (and possibly some other environments where file protection API is not available).

Without this additional check, when the file logger is used in the simulator in an app that can run in the background, the log file is always rotated at launch regardless how large or how old it is (because both key comparisons here return false, so YES is returned from this code block).


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

ffried:
@mackuba Thanks for the PR. Could you quickly rebase it (some files were renamed due to #1083)? I've already added the change log entry on `master` for this change.

mackuba:
Rebased 👍

ffried:
@mackuba Thank you! I'll merge this as soon as Travis is done...

PR #1099, opened 2019-09-30
Prepare release 3.6.0

ffried:
This updates the projects to Xcode 11 and adds some missing changelog entries.
It further prepares the release of a version 3.6.0

PR #1100, opened 2019-09-30
Limit assertion to non-simulator build

ffried:

### Pull Request Description

We have an assertion that checks whether the parent folder exists or not. This might fail in simulator builds where we move the log files to archive them, since the parent folder might not exist anymore in this case.
This fixes that problem by limiting the assertion to non-simulator builds.


ffried:
(I've added the changelog entry in #1099 to avoid conflicts)

PR #1118, opened 2020-01-22
Fixed setxattr() function usage.

jcbertin:

### Pull Request Description

This should fix extended attributes in simulator.


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Warnings
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1673,10 +1508,10 @@ - (BOOL)hasExtendedAttributeWithName:(NSString *)attrName {
 }
 
 - (void)addExtendedAttributeWithName:(NSString *)attrName {
-    const char *path = [filePath UTF8String];
+    const char *path = [filePath fileSystemRepresentation];
     const char *name = [attrName UTF8String];
 
-    int result = setxattr(path, name, NULL, 0, 0, 0);
+    int result = setxattr(path, name, "", 1, 0, 0);
```
ffried: I think we should put a "1" or "true" in there. For any program that simply reads extended attributes, it'll make it easier to understand our attribute.

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1526,145 +1498,8 @@ - (void)renameFile:(NSString *)newFileName {
...
-}
-
-#else /* if TARGET_IPHONE_SIMULATOR */
-
 - (BOOL)hasExtendedAttributeWithName:(NSString *)attrName {
-    const char *path = [filePath UTF8String];
+    const char *path = [filePath fileSystemRepresentation];
     const char *name = [attrName UTF8String];
 
     ssize_t result = getxattr(path, name, NULL, 0, 0, 0);
```
ffried: With below's comment about "1" or "true", I think we should also check this here. We just need to be careful about backwards compatibility.

ffried: To maintain backwards compatibility, we could go ahead and have a fast-path, that checks for the value we set (e.g. "1"). If it doesn't match, we can (for now) fall back to the old check, but replace the extended attribute with the new value, so that next time the first check will succeed. With that in place, we can drop legacy code altogether in some future version (whereas now we'd still have to keep it).

Sources/CocoaLumberjack/DDFileLogger.m
jcbertin: Fixed in 66a88e0.

Sources/CocoaLumberjack/DDFileLogger.m
jcbertin: Fixed in 9e9c7b2

jcbertin:
Rebased on master.

Sources/CocoaLumberjack/include/DDFileLogger.h
```diff
@@ -534,22 +534,15 @@ extern unsigned long long const kDDDefaultLogFilesDiskQuota;
 // "mylog"     -> "mylog.archived"
```
ffried: I think we should adjust the comments here. Maybe even remove this API from the public header, since it should no longer be used by clients.

jcbertin: BTW, should we keep -renameFile: method in the public API?

ffried: I don't think they do any harm, so yea. `- reset` will simply reset the lazy vars, which is fine. `-renameFile:` is also not dangerous, as long as the user is aware that this might create a new log file - which might actually be what the user wants. Imagine someone wanting to "mark" a log file by first rolling the log file and then renaming the old one. This would also potentially "save" the log file from being removed at a later point.

ffried:
Looks good now. As soon as the CI is through, I'll merge this. Thanks a lot @jcbertin!

PR #1111, opened 2019-11-14
Deal with errors during log file creation

ffried:
### New Pull Request Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](http://cocoadocs.org/docsets/CocoaLumberjack/)
* [x] I have searched for a similar pull request in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/pulls) and found none

* [x] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)
* [] I have added the required tests to prove the fix/feature I am adding
* [x] I have updated the documentation (if necessary)
* [x] I have run the tests and they pass
* [x] I have run the lint and it passes (`pod lib lint`)

This merge request fixes / refers to the following issues: #1103 

### Pull Request Description

As outlined in #1103, we might return nil from `-[DDLogFileManagerDefault createNewLogFile]` if critical errors occurred a few times in a row. 
I've changed the API now to allow propagating the errors (since `DDLogFileManager` is a protocol) and deal with them in `DDFileLogger`. Still, the nullability annotation of `currentLogFileInfo` had to change to `nullable` since we have to return null there, if the creation of a log file fails.

PR #1119, opened 2020-01-23
Improve nullability annotations for Swift

ffried:

This merge request fixes / refers to the following issues: -

### Pull Request Description

This improves the nullability annotations in many header files we have. Note that #1111 also adds some of these on `DDLogFileInfo`, which is why I didn't change those here.


OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Sources/CocoaLumberjack/DDLog.m
```diff
@@ -1242,9 +1242,7 @@ - (void)setLogFormatter:(id <DDLogFormatter>)logFormatter {
         }
     };
 
-    dispatch_queue_t globalLoggingQueue = [DDLog loggingQueue];
-
-    dispatch_async(globalLoggingQueue, ^{
+    dispatch_async([DDLog loggingQueue], ^{
```
lolgear: Nowadays `DDLog.loggingQueue` is more "idiomatic" in terms of interoperability/readability, no? 
"Shape of Swift", hah.

ffried:
> Do we need remove if let sentence because now NS_ASSUME_NONNULL has been added in DDOSLogger.h?

@sushichop Hah, you're right! Missed the integration targets. Thanks!

Sources/CocoaLumberjack/DDLog.m
ffried: Done.

Sources/CocoaLumberjack/DDLog.m
```diff
@@ -705,21 +703,18 @@ - (void)lt_addLogger:(id <DDLogger>)logger level:(DDLogLevel)level {
...
 
     if (loggerQueue == nil) {
         // Automatically create queue for the logger.
         // Use the logger name as the queue name if possible.
-
         const char *loggerQueueName = NULL;
 
         if ([logger respondsToSelector:@selector(loggerName)]) {
-            loggerQueueName = [[logger loggerName] UTF8String];
+            loggerQueueName = logger.loggerName.UTF8String;
```
lolgear: Should we add logger.loggerQueueName property here?

ffried: Why? What's wrong with having two dots? 😉

ffried: Also, where would you add that property? You can't add a property with a default implementation on a protocol in ObjC 😉

lolgear: @ffried I could wrap everything in `NSProxy` ... :D
Or we could add static function that get queue name of a logger.
Nothing wrong with two dots, yes, but I would like to distinguish `loggerName property` and `loggerQueueName transform`.
Or it could be done via comment in logger protocol.

```objective-c
// NOTE: We use loggerName .utf8 representation to set loggerQueueName.
// @property (copy, nonatomic, readonly) char *loggerQueueName; // logger.loggerName.UTF8String;
```

ffried: I don't think that any action is necessary here. We never had a problem with it and it's an internal matter what we do with `loggerName`. Especially since it's an extensible enum, so we basically should always have compile time constants here, which means we also always get an UTF-8 string.

Since we don't need to transform it very often, I think we're fine with what we have now.

lolgear: @ffried 
Just to be sure.
Does `loggerName` allow emoji?

ffried: Allowed is any string the compiler can process as constant. 
But even emojis are representable in UTF-8 using surrogate pairs:

```objc
#import <Foundation/Foundation.h>

typedef NSString *LoggerName NS_TYPED_EXTENSIBLE_ENUM;

static LoggerName const LoggerNameEmoji = @"🤷‍♂️🧔🏽";

int main(int argc, char *argv[]) {
	@autoreleasepool {
		dispatch_queue_t q = dispatch_queue_create(LoggerNameEmoji.UTF8String, NULL);
		const char *label = dispatch_queue_get_label(q);
		NSString *strLabel = [[NSString alloc] initWithUTF8String:label];
		NSLog(@"%@", strLabel); // -> 2020-01-23 18:27:19.651 Untitled[62409:8550421] 🤷‍♂️🧔🏽
	}
}
```

PR #1121, opened 2020-01-24
NSDateFormatter thread safety

jcbertin:

### Pull Request Description

NSDateFormatter is thread safe since iOS 7 and macOS 10.9 (cf. https://developer.apple.com/documentation/foundation/nsdateformatter). So we could remove thread safety code when using NSDateFormatter. My only concern is watchOS and tvOS: I could not find any information about thread safety on those platforms.
Sideways, this PR fixes DDAtomicCounter not being monotonic. One question is: should we keep this public API since it was only use by NSDateFormatter thread safety code?


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - Fix rolling timer being rescheduled rapidly due to leeway (#1106 / #1107)
 - Fix -didArchiveLogFile: returning the file name instead of the file path (#1078)
 - Fix setxattr() function usage
+- Fix NSDateFormatter thread safety
```
ffried: Btw. please add the pull request ID to the changelog (maybe also for your last PR, since I missed it there 😅). Makes it easier to get to the full discussion (see entries above). 🙂

OSSPolice:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

jcbertin:
Good point as usual! 😉

CHANGELOG.md
jcbertin: See 860c578

lolgear:
@ffried @jcbertin 
We could remove these deprecated symbols in 4.0.0... 
What do you think about it?

Sources/CocoaLumberjack/include/DDDispatchQueueLogFormatter.h
```diff
@@ -27,6 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Log formatter mode
  */
+__attribute((deprecated))
```
ffried: There is a nice (modern) macro `API_DEPRECATED` to declare deprecations, which also takes a message. 😉

Sources/CocoaLumberjack/include/DDDispatchQueueLogFormatter.h
```diff
@@ -91,7 +92,7 @@ typedef NS_ENUM(NSUInteger, DDDispatchQueueLogFormatterMode){
  *
  *  @param mode choose between DDDispatchQueueLogFormatterModeShareble and DDDispatchQueueLogFormatterModeNonShareble, depending if the formatter is shared between several loggers or not
  */
-- (instancetype)initWithMode:(DDDispatchQueueLogFormatterMode)mode;
+- (instancetype)initWithMode:(DDDispatchQueueLogFormatterMode)mode __attribute((deprecated));
```
ffried: As above: `API_DEPRECATED`

Sources/CocoaLumberjack/include/DDDispatchQueueLogFormatter.h
```diff
@@ -178,8 +179,9 @@ typedef NS_ENUM(NSUInteger, DDDispatchQueueLogFormatterMode){
 
 @end
 
-#pragma mark - DDAtomicCounter
+#pragma mark - DDAtomicCountable
 
+__attribute((deprecated))
```
ffried: Last but not least again here: `API_DEPRECATED`

Sources/CocoaLumberjack/include/DDDispatchQueueLogFormatter.h
```diff
@@ -189,7 +191,4 @@ typedef NS_ENUM(NSUInteger, DDDispatchQueueLogFormatterMode){
 
 @end
 
-@interface DDAtomicCounter: NSObject<DDAtomicCountable>
```
ffried: Oh, that one was public? 😬
Then we have to keep it but mark it as deprecated as well. Could have sworn it was only available internally...

Sources/CocoaLumberjack/include/DDDispatchQueueLogFormatter.h
jcbertin: I don't think we can use that here: `API_DEPRECATED` needs platform version parameter which is of no use here.
We can use `__attribute((deprecated))` with message too but I'm wondering which message to use.

Sources/CocoaLumberjack/include/DDDispatchQueueLogFormatter.h
jcbertin: See 6ee355f

Sources/CocoaLumberjack/include/DDDispatchQueueLogFormatter.h
ffried: Ah, you're right. That's what I get for only doing a quick repo search instead of looking at the source code. GitHub found a usage of API_DEPRECATED where the platforms were on the next line, so I thought it could be used without platforms. 🤦‍♂️

A message would still be nice, but I see you already added those. Thanks!

PR #1124, opened 2020-01-31
Added support for thread QOS in DDLogMessage class.

jcbertin:

### Pull Request Description

IMHO this is a good addition that helps a lot when debugging threads stuff.
What do you think about it?

### DIFF
--- a/Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
+++ b/Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
@@ -221,7 +221,9 @@ static const char* _qos_string(NSUInteger qos) {
     NSString *timestamp = [self stringFromDate:(logMessage->_timestamp)];
     NSString *queueThreadLabel = [self queueThreadLabelForLogMessage:logMessage];
 
-    return [NSString stringWithFormat:@"%@ [%@ (QOS:%s)] %@", timestamp, queueThreadLabel, _qos_string(logMessage->_qos), logMessage->_message];
+    if (@available(macOS 10.10, iOS 8.0, *))
+        return [NSString stringWithFormat:@"%@ [%@ (QOS:%s)] %@", timestamp, queueThreadLabel, _qos_string(logMessage->_qos), logMessage->_message];
+    return [NSString stringWithFormat:@"%@ [%@] %@", timestamp, queueThreadLabel, logMessage->_message];
 }
 
 @end
--- a/Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
+++ b/Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
@@ -16,10 +16,22 @@
 #import <CocoaLumberjack/DDDispatchQueueLogFormatter.h>
 #import <pthread/pthread.h>
 #import <stdatomic.h>
+#import <sys/qos.h>
 
 #if !__has_feature(objc_arc)
 #error This file must be compiled with ARC. Use -fobjc-arc flag (or convert project to ARC).
 #endif
+
+static const char* _qos_string(NSUInteger qos) {
+    switch ((qos_class_t) qos) {
+        case QOS_CLASS_USER_INTERACTIVE: return "UI";
+        case QOS_CLASS_USER_INITIATED:   return "IN";
+        case QOS_CLASS_DEFAULT:          return "DF";
+        case QOS_CLASS_UTILITY:          return "UT";
+        case QOS_CLASS_BACKGROUND:       return "BG";
+        default:                         return "UN";
+    }
+}
 
 #pragma mark - DDDispatchQueueLogFormatter
 
@@ -209,7 +221,7 @@
     NSString *timestamp = [self stringFromDate:(logMessage->_timestamp)];
     NSString *queueThreadLabel = [self queueThreadLabelForLogMessage:logMessage];
 
-    return [NSString stringWithFormat:@"%@ [%@] %@", timestamp, queueThreadLabel, logMessage->_message];
+    return [NSString stringWithFormat:@"%@ [%@ (QOS:%s)] %@", timestamp, queueThreadLabel, _qos_string(logMessage->_qos), logMessage->_message];
 }
 
 @end
--- a/Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
+++ b/Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
@@ -22,14 +22,21 @@
 #error This file must be compiled with ARC. Use -fobjc-arc flag (or convert project to ARC).
 #endif
 
-static const char* _qos_string(NSUInteger qos) {
+DDQualityOfServiceName const DDQualityOfServiceUserInteractive = @"UI";
+DDQualityOfServiceName const DDQualityOfServiceUserInitiated   = @"IN";
+DDQualityOfServiceName const DDQualityOfServiceDefault         = @"DF";
+DDQualityOfServiceName const DDQualityOfServiceUtility         = @"UT";
+DDQualityOfServiceName const DDQualityOfServiceBackground      = @"BG";
+DDQualityOfServiceName const DDQualityOfServiceUnspecified     = @"UN";
+
+static DDQualityOfServiceName _qos_name(NSUInteger qos) {
     switch ((qos_class_t) qos) {
-        case QOS_CLASS_USER_INTERACTIVE: return "UI";
-        case QOS_CLASS_USER_INITIATED:   return "IN";
-        case QOS_CLASS_DEFAULT:          return "DF";
-        case QOS_CLASS_UTILITY:          return "UT";
-        case QOS_CLASS_BACKGROUND:       return "BG";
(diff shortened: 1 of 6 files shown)


Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
```diff
@@ -16,11 +16,23 @@
...
 #import <stdatomic.h>
+#import <sys/qos.h>
 
 #if !__has_feature(objc_arc)
 #error This file must be compiled with ARC. Use -fobjc-arc flag (or convert project to ARC).
 #endif
 
+static const char* _qos_string(NSUInteger qos) {
+    switch ((qos_class_t) qos) {
+        case QOS_CLASS_USER_INTERACTIVE: return "UI";
```
lolgear: What is a purpose to corrupt name?
Could consumer change default values?
Why method name is _qos_string and return values are "qos_codes"?

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
```diff
@@ -209,7 +221,7 @@ - (NSString *)formatLogMessage:(DDLogMessage *)logMessage {
     NSString *timestamp = [self stringFromDate:(logMessage->_timestamp)];
     NSString *queueThreadLabel = [self queueThreadLabelForLogMessage:logMessage];
 
-    return [NSString stringWithFormat:@"%@ [%@] %@", timestamp, queueThreadLabel, logMessage->_message];
+    return [NSString stringWithFormat:@"%@ [%@ (QOS:%s)] %@", timestamp, queueThreadLabel, _qos_string(logMessage->_qos), logMessage->_message];
```
lolgear: Could person opt-out of these "qos-codes"? ( for example, somebody may have parsers for logs... )

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
jcbertin: Corrupted names? You mean returning "UI" instead of "USER_INTERACTIVE" for example?
_qos_string() is returning C string hence it's name.

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
ffried: You shouldn't parse logs that aren't formatted by your code. We don't guarantee a steady log format (neither does Apple btw)

lolgear: @ffried Sad :(

jcbertin: We could add options property for DDDispatchQueueLogFormatter class. And default it by not printing QOS…

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
lolgear: @jcbertin 
We have to do something with these codes.
`UN/UT/UI/IN` - not ok.

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
ffried: What we should definitively do is to not print it if it's not available.
But otherwise I'd have no problem with simply adding it. Like I said, Apple does the same.
You can always run your own formatter that does not print it. Especially since we don't use this formatter by default.

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
jcbertin: I used those code because they are small and use extensively in WWDC keynotes. Do you want me to change it to longer version anyway?

ffried: I agree with @lolgear that those codes are a little "cryptic". AFAIK, Apple also shows the full name in the Xcode sidebar.

ffried: Interesting. Apple uses them as well? If so, I'd have not problem using them.

Sources/CocoaLumberjack/include/DDLog.h
```diff
@@ -850,6 +851,7 @@ typedef NS_OPTIONS(NSInteger, DDLogMessageOptions){
 @property (readonly, nonatomic) NSString *threadID; // ID as it appears in NSLog calculated from the machThreadID
 @property (readonly, nonatomic, nullable) NSString *threadName;
 @property (readonly, nonatomic) NSString *queueLabel;
+@property (readonly, nonatomic) NSUInteger qos;
```
ffried: Should we mark this as available as well?

jcbertin: Good point!

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
jcbertin: See https://developer.apple.com/videos/play/wwdc2015/718/

lolgear: Time 10:11.

Interesting. Maybe we could, well... add "emoji" to distinct qos values?

Sources/CocoaLumberjack/include/DDLog.h
jcbertin: See 3a35cc7

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
jcbertin: See 3a35cc7

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
jcbertin: 😋

lolgear: @ffried @jcbertin 
O!
We could save codes, but for that we should somehow tell consumer what they mean.

```objective-c
extern NSString *CCQOSStringUserInteractive= "UI";
```

What do you think?

jcbertin: @lolgear yes it could clarify the meaning of those codes.

ffried: I don't think that this should be our responsibility. If Apple uses those, they should provide them as well. So this would be something for feedbackassistant.apple.com 😉

Sources/CocoaLumberjack/include/DDLog.h
```diff
@@ -850,6 +851,7 @@ typedef NS_OPTIONS(NSInteger, DDLogMessageOptions){
 @property (readonly, nonatomic) NSString *threadID; // ID as it appears in NSLog calculated from the machThreadID
 @property (readonly, nonatomic, nullable) NSString *threadName;
 @property (readonly, nonatomic) NSString *queueLabel;
+@property (readonly, nonatomic) NSUInteger qos API_AVAILABLE(macos(10.10), ios(8.0));
```
ffried: We do also support watchOS and tvOS. So those availability marks should be present as well (also in the `if (@available(...))` statements.

Sources/CocoaLumberjack/Extensions/DDDispatchQueueLogFormatter.m
jcbertin: See c000783

Sources/CocoaLumberjack/include/DDLog.h
jcbertin: Yes but qos_class_self() doesn't define availability for those OSes…

ffried: Fair point.

Sources/CocoaLumberjack/include/DDDispatchQueueLogFormatter.h
```diff
@@ -41,6 +41,36 @@ typedef NS_ENUM(NSUInteger, DDDispatchQueueLogFormatterMode){
...
+ * Since macOS 10.10 and iOS 8.0, pthreads, dispatch queues and NSOperations express their
+ * scheduling priority by using an abstract classification called Quality of Service (QOS).
+ *
+ * This formatter will add a representation of this QOS in the log message by using those
+ * string constants.
+ * For example:
+ *
+ * `2011-10-17 20:21:45.435 AppName[19928:5207 (QOS:DF)] Your log message here`
+ *
+ * Where QOS is one of:
```
lolgear: Great!

jcbertin: My pleasure!

jcbertin:
Yeah! Merge time…

PR #1135, opened 2020-04-23
Port Travis configuration to GitHub actions

ffried:
### New Pull Request Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](http://cocoadocs.org/docsets/CocoaLumberjack/)
* [x] I have searched for a similar pull request in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/pulls) and found none

* [x] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)
* [x] I have added the required tests to prove the fix/feature I am adding
* [x] I have updated the documentation (if necessary)
* [x] I have run the tests and they pass
* [x] I have run the lint and it passes (`pod lib lint`)

This merge request fixes / refers to the following issues: -

### Pull Request Description

This attempts to ditch Travis and use GitHub Actions instead. Travis seems to have become slow in adopting new OS' (e.g. there's still no macOS 10.15).

A few things to note:
- I've used [our own action](https://github.com/sersoft-gmbh/xcodebuild-action) for running `xcodebuild` commands. I found this easier to maintain, but feel like it's worth mentioning since I'm also developing said action. I'd have no issue with converting back to manual command execution if you guys prefer that.
- I've split the tasks into multiple workflow files. This makes it easier to see what is what.
- This PR is still WIP since I need to test whether a few things work. Also, I'll eventually need the support of one of the repo admin to make the final switch (setting required PR checks, maybe add the codecov token to secrets).

PR #1140, opened 2020-04-27
Migrate to DangerSwift and use Danger's own GitHub action

ffried:

This merge request fixes / refers to the following issues: -

### Pull Request Description

Apparently Danger has problem with running when a PR is made from a fork after using GitHub Actions. There is an issue about this in the danger repo [here](https://github.com/danger/danger/issues/1103).
However, since there is also DangerSwift, I've migrated away from ruby Danger. Since Swift is probably easier to understand for users / collaborators of this repository 😉 

Also: run Danger on Ubuntu - it does not need macOS


lolgear:
@ffried 
What is the reason to switch to Github Actions? 
Just curious about it.

ffried:
@lolgear See #1135
In a nutshell: Travis became slow at supporting new stuff. Also: Why use a 3rd party service if GitHub can do the same (or actually even more)? 😉

ffried:
@sushichop Yea, however Danger still fails for PRs originating from forks. I'm not sure if we used a personal access token for the old danger integration. And tbh, I'd prefer to use the GitHub provided token if possible. There are still discussions going on in GitHubs forums about this.

sushichop:
@ffried  Oops! I missed that. Sorry. This PR basically seemed to be OK...😅

ffried:
No worries! It's why the PR is still marked WIP. I haven't found time yet to take another look at it.

ffried:
@bpoplauschi I think I need your help with this again. Apparently, the access token provided by GitHub is not enough for running danger on a pull request coming from a fork. There are various forums, blogs, ... about this (just search for "GitHub actions token pull request fork").

Most people recommend using a personal access token (PAT) for this. I think we had this before, when we were using travis, right? At least the user posting the comments looked like a bot-account with a PAT. We'd need a PAT inside a secret variable named `DANGER_GITHUB_API_TOKEN`. Could you add that to the repo for me? I'd re-run the checks then to see if that works.

bpoplauschi:
@ffried it took me a while to remember this, but you are correct. 
When I initially setup Danger, I created the https://github.com/OSSPolice account, then created a PAT under that account and created an env var inside the Travis-Ci repo config named exactly `DANGER_GITHUB_API_TOKEN`. 
After I found that token, I now created a repository secret (just for this repo) with the name name and that value.
For now, you can just rerun the checks. If you need it, we can discuss about giving you more permissions.

ffried:
@bpoplauschi Thanks a lot! If only that would have solved the issue... 😞 

I've re-checked the sources I've found about using a PAT and they're not too old, but old enough that the solution no longer works. The underlying problem is, that GitHub Actions does *not* pass secrets to actions triggered by forks. I *do* understand the security reasons behind that decision (someone could easily steal secrets). I *do not* understand, however, that they still haven't come up with a solution to this. There are many possible and likely easy fixes - like making it a repo-owner decision, or like providing a more restricted token (with e.g. only access to the PR which triggered the action).

But these are all just nice ideas - no solutions. In the current situation, we cannot run Danger for PRs opened from a fork of our repository. There are a couple of workarounds, but only two seem to work reliably:
- As it seems, uploading resources seems to be allowed with the GitHub-provided token (not the PAT) though. One workaround creates temporary files for each opened PR and a cron-like GitHub action (that is then triggered from our repo which has the token with enough rights) then works these files and would then run danger.
- We opt out of running danger automatically for PRs originating from forks for now, and instead add a slash-command for running danger. Since slash-commands are triggered by comment-actions which are also executed in our repo, these would get the token with enough rights as well..

IMHO, option 1 is too much hassle. Option 2 would be fairly simple to implement and thus also fairly simple to undo once GitHub comes up with a solution to this. It would require more work from our side, though, since danger no longer runs fully automated but we'd have to add a slash-command. Partially defeats the purpose of danger, but is still better than manually remembering all the things danger checks.

Btw., automatically adding the "danger-comment" doesn't work either, because of the same reason: No token has enough rights to do this.

@CocoaLumberjack/collaborators What do you think?

bpoplauschi:
@ffried wow, this is a lot deeper issue than I thought. IMO, we should contact GitHub support about this, maybe they can give us more info / suggestions. And I think we can temp go with option 2 for now.
Let's see what other contributors think too :)

ffried:
@bpoplauschi I think GitHub support is aware of the issue:
- https://github.community/t/github-workflow-not-running-from-pull-request-from-forked-repository
- https://github.community/t/github-action-workflow-is-executed-for-a-pr-from-a-forked-repo-if-the-pr-changes-the-workflow-file
- And many smaller ones spread in the community forums.

DangerJS has similar issues: https://github.com/danger/danger-js/issues/918
I've seen issues in other Repos as well. One of the repos I've seen have disabled all writing actions for PRs from forks.

sushichop:
@ffried cc: @bpoplauschi 
Thanks for your sharing idea. I also investigated these days about GitHub Actions from fork repo. And my conclusion is the same as you unfortunately... GitHub Actions only grants READ permission even when we use the special secrets(GITHUB_TOKEN). https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#permissions-for-the-github_token


Then I prefer option 2 to option 1 for now. (Is slash command [here](https://github.com/marketplace/actions/slash-commands)? 



> * We opt out of running danger automatically for PRs originating from forks for now, and instead add a slash-command for running danger. Since slash-commands are triggered by comment-actions which are also executed in our repo, these would get the token with enough rights as well..

Or can we choose option 3? But this is just a workaround and go back a little..
- We only use Travis-CI **only for** danger/swift and use GitHub Actions for everything else.

IMO, it' OK with option 2 or 3 for now.

ffried:
@sushichop I thought about using Travis as well, but I'd like to keep DangerSwift since contributors to this repository probably have better experience in Swift than in other languages. However, using DangerSwift outside of their GitHub action currently requires having Danger as SPM dependency - which would probably confuse users of CocoaLumberjack integrating via SPM.

I've remodeled the validation workflow to also trigger on comment events and listen for `/run_checks` to run Danger. Only members with write-access to the repo can trigger it. Testing is difficult, however, since `issue_comment` workflows only trigger when their workflow file is in `master`...

Also, I've kept the automatic run of Danger for internal PRs.

sushichop:
@ffried 

> I thought about using Travis as well, but I'd like to keep DangerSwift since contributors to this repository probably have better experience in Swift than in other languages. However, using DangerSwift outside of their GitHub action currently requires having Danger as SPM dependency - which would probably confuse users of CocoaLumberjack integrating via SPM.

Yeah, that makes sense.

> I've remodeled the validation workflow to also trigger on comment events and listen for `/run_checks` to run Danger. Only members with write-access to the repo can trigger it. Testing is difficult, however, since `issue_comment` workflows only trigger when their workflow file is in `master`...
> 
> Also, I've kept the automatic run of Danger for internal PRs.

Awesome👍
Now I understand CI flow deeply. I totally agree with option 2.

ffried:
@bpoplauschi I think you need to merge this PR with "admin magic powers", since it currently cannot pass the required danger check...

Speaking of which: We might need to un-mark the danger checks as required. Since only one can ever run, we cannot mark both as required...

bpoplauschi:
@ffried sorry it took me so long to look into this, I'm merging the PR now.

ffried:
@bpoplauschi No worries. Thank you! I'll test things tomorrow.

Is GitHub able to specify that one of two checks are required? If so we could still make the Danger check required. Otherwise we need to remove its required flag for reasoned explained above.

bpoplauschi:
> Is GitHub able to specify that one of two checks are required? If so we could still make the Danger check required. Otherwise we need to remove its required flag for reasoned explained above.

I have no idea :)

PR #1148, opened 2020-07-03
SPM: Bump tools-version to 5.3 (keeping 5.2 support)

martinpucik:

This merge request fixes / refers to the following issues: 
- None

### Pull Request Description

iOS version 8.0 is no longer supported as the minimum deployment version with Swift 5.3 and Xcode 12. Adding CocoaLumberjack to a project as a dependency via SPM will produce this warning in Xcode 12:

```
The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.0.99.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Public
 
 - **Breaking change**: Dropped support for iOS 8 (#1153)
+- Update SPM tools-version to 5.3 to enable Swift 5.3 support (#1148)
 
 ## [3.6.2 - Xcode 11.6 on July 31st, 2020](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.6.2)
 
--- a/Package@swift-5.2.swift
+++ b/Package@swift-5.2.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.2
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription


ffried:
@martinpucik Thanks for your contribution! However, while Swift 5.3 is still in beta (alongside Xcode 12), I'll mark this as draft. Also, we should probably coordinate updating our Xcode project together with the SPM project file.

Kaspik:
@ffried As we are waiting for new release from master with many fixes, should there be a branch `swift_5_3` that we can merge this to and open more PRs that will be merged to, but still be able to use this branch by users using XCode 12? It would help a lot.

When Swift 5.3 and XCode 12 will be out, that branch would be merged to master. :)

wdyt?

ffried:
@Kaspik AFAIK, there aren't currently any breaking issues with CocoaLumberjack and Xcode 12, are there? Xcode will automatically cap the deployment target to 9.0 in Xcode 12 - emitting above's warning. IMHO warnings are to be expected in betas - as long as the rest works fine they aren't a big deal.
Or am I missing something here?

Kaspik:
I haven't tried opening the project itself on XCode 12 yet, but I think there will be some small update from Swift 5.2 to Swift 5.3 in the codebase as well? Great if not.

Having a branch with Swift 5.3 default support now could be just an option for XCode 12 users to _get rid of the warning_ (we are trying to make it warning free for iOS 14 before XCode 12 comes out, better than after it comes out), possibly make other changes and also be ready in advance for you so you have less work once it's released.

Is there a problem with having _beta branch_? Of course ignore me if there is 🙂

But even non XCode 12 project throws warning now (https://github.com/CocoaLumberjack/CocoaLumberjack/issues/1149 - which you are aware of), so I would be happy to get both of them fixed. 🚀

ffried:
@Kaspik I've had the project open in Xcode 12 and things build find (even on that new kind of CPU for macOS no one is allowed to talk about 😉).

I think our goal should be to be ready somewhat alongside the GM of iOS 14 and Xcode 12. Before that, there is too much fluctuation in what will make it into the release and what not. Our Swift codebase is not that big, most of the code is still ObjC. The only thing I think will affect us is the `#filePath` vs. `#file` topic - which is exactly one of thee topics where the Swift team is still unsure how it'll end up in the final 5.3 release: 
- https://github.com/apple/swift-evolution/blob/master/proposals/0274-magic-file.md
- https://github.com/apple/swift-evolution/blob/master/proposals/0285-ease-pound-file-transition.md

Given that there aren't currently any breaking changes with Xcode 12, I'd prefer to wait until a few more beta's have seen the light of the world. Looking at the last few years, things tend to get more stable around beta 6. By then I think we can consider merging such changes directly to `master`. Maybe we could release 3.6.2 before that, so that we have a small bugfix version out with the current fixes on `master`.

Long story short: As long as we don't have any major problems with Xcode 12 / any new OS (which is currently the case), I'd prefer not to have a separate branch (that we'd need to keep up to date with master).


Apart from that, your points on what we have to do are correct. From the top of my head I don't think that much will change, though. Most our availability checks aim above iOS 9 and apart from above's `#filePath` / `#file`  topic, our Swift code will likely stay the same as well...

PR #1153, opened 2020-07-24
Drop iOS 8, remove unnecessary code

Kaspik:

This merge request fixes / refers to the following issues: none

### Pull Request Description
As per discussion in https://github.com/CocoaLumberjack/CocoaLumberjack/pull/1148 with @ffried - soon iOS 8 won't be available in Xcode (in Xcode 12).

This PR prepares CocoaLumberJack for upcoming release of iOS and Xcode and simply just drops iOS 9.
Following PRs in future will also update SPM and project itself to Swift 5.3.
- Drop iOS 8 in order to prepare for September and iOS 14 with Xcode 12 and his minimum deployment target as iOS 8
- Remove availability macros that are not necessary as CocoaLumberJack supports macOS 10.10 and iOS 9.0 and higher

This fixed warnings for Xcode 12 without updating the codebase and Swift version itself and without affecting master and stable Xcode itself - which sounds like a good step to me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [3.6.2 - Xcode 11.4 on ???, 2020](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.6.2)
 
 ### Public
+- **Breaking change**: Dropped support for iOS 8 (#1153)
 - Fix warnings when building with SPM bundled with Swift 5.2 / Xcode 11.4 (#1132)
 - Added Swift name for DDQualityOfServiceName constants.
 - Don't localize timestamps in `DDefaultFileLogFormatter` (#1151)
--- a/CocoaLumberjack.podspec
+++ b/CocoaLumberjack.podspec
@@ -18,7 +18,7 @@ Pod::Spec.new do |s|
 
   s.preserve_paths = 'README.md'
 
-  s.ios.deployment_target     = '8.0'
+  s.ios.deployment_target     = '9.0'
   s.osx.deployment_target     = '10.10'
   s.watchos.deployment_target = '3.0'
   s.tvos.deployment_target    = '9.0'
--- a/Configs/Module-Shared.xcconfig
+++ b/Configs/Module-Shared.xcconfig
@@ -189,7 +189,7 @@ GCC_WARN_UNUSED_VARIABLE = YES
 HEADER_SEARCH_PATHS = $(inherited)
 
 // Code will load on this and later versions of iOS. Framework APIs that are unavailable in earlier versions will be weak-linked; your code should check for null function pointers or specific system versions before calling newer APIs.
-IPHONEOS_DEPLOYMENT_TARGET = 8.0
+IPHONEOS_DEPLOYMENT_TARGET = 9.0
 
 // This is a list of paths to be added to the `runpath` search path list for the image being created. At runtime, `dyld` uses the `runpath` when searching for dylibs whose load path begins with `@rpath/`.
 LD_RUNPATH_SEARCH_PATHS = $(inherited) @executable_path/Frameworks @loader_path/Frameworks
--- a/Package.swift
+++ b/Package.swift
@@ -6,7 +6,7 @@ import PackageDescription
 let package = Package(
     name: "CocoaLumberjack",
     platforms: [
-        .iOS(.v8),
+        .iOS(.v9),
         .macOS(.v10_10),
         .watchOS(.v3),
         .tvOS(.v9),
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Then use `DDOSLogger` for iOS 10 and later, or `DDTTYLogger` and `DDASLLogger` f
 #### CocoaPods
 
 ```ruby
-platform :ios, '8.0'
+platform :ios, '9.0'
 
 target 'SampleTarget' do
   use_frameworks!
@@ -35,7 +35,7 @@ For more details about how to use Swift with Lumberjack, see [this conversation]
 
 For Objective-C use the following:
 ```ruby
-platform :ios, '8.0'
+platform :ios, '9.0'
 
 target 'SampleTarget' do
     pod 'CocoaLumberjack'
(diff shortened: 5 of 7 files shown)


Kaspik:
Screen Shot 2020-07-24 at 17 37 46

I wonder what's going on with Codecov. Did the project coverage really increase that much couple of months ago? And why it decreased now in this PR - by removing 4 lines? 😂 

Re: Carthage - I'm not sure how to update it? Can't seem to find the Carthage file...

ffried:
@Kaspik I have no idea what's going on with codecov recently. Noticed that as well but ¯\_(ツ)_/¯

The warning about Carthage can be ignored. You (correctly) made changes to the podspec but did not touch the Xcode project - which is also correct, since you updated the xcconfigs. I might need to refine that check to not emit a warning if the xcconfig files were touched.

Regarding tests: I _think_ we can ignore that warning, too, but could you double check that we don't have any tests for the QoS stuff? Don't know that by heart right now.

Also: as mentioned in #1148 I'm not sure whether this change will/should make it into 3.6.2 or if this is rather something for 3.6.3 (or 3.7?) because of it's "breaking-ness". I doubt that too many apps will be affected - given that all apps that update to Xcode 12 are forced to drop iOS 8 - but it's just a thought.

So if you don't mind, I'd like keep this open for now and see if we can decide in the @CocoaLumberjack/collaborators team on what should go into what release and when to release them...

ffried:
@Kaspik Thanks for taking this on, btw! Greatly appreciated!

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## [3.6.2 - Xcode 11.4 on ???, 2020](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.6.2)
 
 ### Public
+- **Breaking change**: Dropped support for iOS 8 (#1153)
```
ffried, line 4: As mentioned: depending on when we release what version, this might need to go into another section... This is merely a reminder note, though. No action needed at the moment 🙂

Kaspik:
Cool, thanks!

I was thinking about marking this as 3.7.0 as well - as current master would be nice to get out for everyone as 3.6.2 - so I'm totally okay with that and I can bump it after 3.6.2 release to merge this. 😊

Kaspik:
@ffried Hey! Any update on the 3.6.2 release please? :)

lolgear:
@Kaspik Drop iOS 9?
I think that Xcode 12 is the last Xcode that will support iOS 9.
Am I right?

Kaspik:
@lolgear Sorry, wrong title - this is dropping iOS 8 that is not supported by XCode 12 anymore.
But 3.6.2 will still support it.

lolgear:
@Kaspik 
@ffried 

As you both mentioned above "breaking changes" and "affected apps".

Do you think that many apps use "correct" versioning?
I guess that people (not all, of course) don't understand which changes affect which number in `semver`.

In my old project I have, for example, `'CocoaLumberjack'` or similar without a number. 
Yeah, I was young :)

So, for me it is fine that something is updating and IF you ( as OSS and library ) drop support for old OS, then, I guess that most users are already beyond this legacy OS.

Again, it is only about iOS. I guess it is different story for macOS.

ffried:
@Kaspik See #1155 

@lolgear SPM does... So does e.g. Carthage. CocoaPods too, I think. They all adhere semver. As a user of a framework you don't need to worry about semver. Xcode (in case of SPM) by default adds a dependency with the restriction "up to next major", since all changes below the next major should be non-breaking.

I still flagged the next release as 3.7.0, since this repo is currently not strictly following semver (not ideal IMHO, but we also never committed to it), and we used a minor increase for the last Xcode update as well.
Since the _actual_ source changes are minimal, and everyone using Xcode 12 will have to drop support for iOS 8, I think it's reasonable.

sushichop:
Hi all, 
in this case, I also think it's a good idea to bump version to 3.7.0 with Xcode12.

ffried:
With 3.6.2 being live, this can no go into master.

ffried:
Ah, shoot. Forgot to move the CHANGELOG entry... See #1156

drademacher:
@ffried I have no idea how to inform you guys, but this was a very bad decision. With adding a breaking change in a minor version you violate the common rules of semantic versioning. (see https://semver.org/ for more)

Other software depends on people making the correct decisions for semantic versioning. As you can see for example in another library (https://github.com/SVGKit/SVGKit/issues/700) this cause the version to be permantently broken. Those existing versions expect no breaking changes in your major version 3, but 3.7 introduces them anyways. So those librarys can not be ever again be downloaded via SPM. Since you are quiet a big library, there will be other projects to be effected by this.

There is no way to fix this, but for all other projects to properly fix this issue. Since if they would go for breaking changes in minor / patch versions, this will cause the next library in the chain to fail.

Please keep in mind that those breaking changes can mess up a lot of projects using SPM.

ffried:
@drademacher I see your point and I've highlighted that two comments ago here: 
> I still flagged the next release as 3.7.0, since this repo is currently not strictly following semver (not ideal IMHO, but we also never committed to it), and we used a minor increase for the last Xcode update as well.

I'm aware of the issues this might be causing with SPM. However, as soon as you're building with Xcode 12, you'll anyways have the issue, since Xcode 12 dropped support for iOS 8.
I think the main mistake here was to drop the support in the Swift 5.2 package file already. This means it'll break with Xcode 11 as well.

I totally agree that semver compliance is something that becomes more and more important. In this particular case, however, it also wouldn't have been the perfect solution.
If we'd released Xcode 12 support as 4.0.0, people using CocoaLumberjack as a dependency from somewhere deeper in the tree might have been unable to upgrade until **all** of their dependencies released new major versions as well. Until then, they would have been stuck with Xcode's telling them that iOS 8 support was removed. So not ideal either...

sushichop:
@drademacher cc: @ffried 
Thanks for your comment.
As ffried says, while we understand the importance of `semver`, it was actually difficult issue to determine for several reasons as above... (There weren't an ideal option...)

PR #1159, opened 2020-08-06
[Proposal] Add `/run_checks` permission

sushichop:

### Pull Request Description

I'd like to limit the users who can execute `/run_checks` to collaborator and admin (I think it was the expected behavior?). I also simplified the action for checking out the pull requests from the fork. In fact, I have already implemented this method in some projects and they works fine.

Do you prefer this proposal? TBH, I'm not sure we really need this feature(limit the users who can execute `/run_checks`), but we can do it😅 and it gets better, I think.

As a reminder,  this change will work fine after the merge(not at pull request) due to GitHub Action issue_event specification.

Thanks.

### DIFF
--- a/.github/workflows/validations.yml
+++ b/.github/workflows/validations.yml
@@ -39,29 +39,13 @@ jobs:
       && github.event.issue.pull_request != null
       && startsWith(github.event.comment.body, '/run_checks')
     steps:
-      - uses: actions/github-script@v2
+      - uses: sushichop/action-repository-permission@v1
         with:
-          script: |
-            github.reactions.createForIssueComment({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              comment_id: context.payload.comment.id,
-              content: "+1"
-            })
-      - uses: actions/github-script@v2
-        id: pr-details
-        with:
-          script: |
-            const { data: pullRequest } = await github.pulls.get({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              pull_number: context.issue.number
-            });
-            return pullRequest;
+          required-permission: write
+          reaction-permitted: '+1'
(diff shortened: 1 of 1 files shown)


.github/workflows/validations.yml
```diff
@@ -39,29 +39,13 @@ jobs:
...
-              pull_number: context.issue.number
-            });
-            return pullRequest;
+          required-permission: write
+          reaction-permitted: '+1'
       - uses: actions/checkout@v2
         with:
-          repository: ${{fromJSON(steps.pr-details.outputs.result).head.repo.full_name}}
-          ref: ${{fromJSON(steps.pr-details.outputs.result).head.ref}}
+          ref: refs/pull/${{ github.event.issue.number }}/head
```
ffried, line 48: I'm not sure that this will work in all situations. I've manually fetched the PR to be independent of any forking situation (e.g. what about 2+ level forks coming back with a PR into upstream).

sushichop: I see. I certainly hadn't thought of that case... So I'd like to revert these line(check out the PR form the fork)

sushichop:
@ffried 
Thanks!  I'm glad to hear your idea! I reverted the lines about checking out the PR from the fork.
Now I've just added execution(`run_check`) limitation. If it's OK, could you merge this?

ffried:
@sushichop After a quick glance at your GitHub action, I saw that you can post a comment if the user does not have the required permissions. Will this happen by default? If not, would you mind adding this to the action? I think that'll help to understand why the checks do not run in certain cases.

sushichop:
@ffried 
Oh, I'm very glad you have checked my actions! It doesn't comment on anything by default.
So I added action definition when a user is not permitted.

PR #1181, opened 2020-11-08
Using new API for `NSFileHandle` to show more infos when has error.

kinarobin:

### Pull Request Description
Some api in `NSFileHandle`  will be **DEPRECATED** , So It's time to update with new api.

### DIFF
--- a/Sources/CocoaLumberjack/DDFileLogger.m
+++ b/Sources/CocoaLumberjack/DDFileLogger.m
@@ -651,8 +651,15 @@ NSTimeInterval     const kDDRollingLeeway              = 1.0;              // 1s
     NSAssert([self isOnInternalLoggerQueue], @"lt_ methods should be on logger queue.");
 
     if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
-        [_currentLogFileHandle synchronizeAndReturnError:nil];
-        [_currentLogFileHandle closeAndReturnError:nil];
+        __autoreleasing NSError *error;
+        BOOL synchronized = [_currentLogFileHandle synchronizeAndReturnError:&error];
+        if (!synchronized) {
+            NSLogError(@"DDFileLogger: Failed to synchronize file: %@", error);
+        }
+        BOOL closed = [_currentLogFileHandle closeAndReturnError:&error];
+        if (!closed) {
+            NSLogError(@"DDFileLogger: Failed to close file: %@", error);
+        }
     } else {
         [_currentLogFileHandle synchronizeFile];
         [_currentLogFileHandle closeFile];
@@ -892,8 +899,15 @@ NSTimeInterval     const kDDRollingLeeway              = 1.0;              // 1s
     }
     
     if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
-        [_currentLogFileHandle synchronizeAndReturnError:nil];
-        [_currentLogFileHandle closeAndReturnError:nil];
+        __autoreleasing NSError *error;
+        BOOL synchronized = [_currentLogFileHandle synchronizeAndReturnError:&error];
+        if (!synchronized) {
+            NSLogError(@"DDFileLogger: Failed to synchronize file: %@", error);
+        }
+        BOOL closed = [_currentLogFileHandle closeAndReturnError:&error];
+        if (!closed) {
+            NSLogError(@"DDFileLogger: Failed to close file: %@", error);
+        }
     } else {
         [_currentLogFileHandle synchronizeFile];
         [_currentLogFileHandle closeFile];
@@ -959,7 +973,11 @@ NSTimeInterval     const kDDRollingLeeway              = 1.0;              // 1s
     if (_maximumFileSize > 0) {
         unsigned long long fileSize;
         if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
-            [_currentLogFileHandle getOffset:&fileSize error:nil];
+            __autoreleasing NSError *error;
+            BOOL succeed = [_currentLogFileHandle getOffset:&fileSize error:&error];
+            if (!succeed) {
+                NSLogError(@"DDFileLogger: Failed to get offset: %@", error);
+            }
         } else {
             fileSize = [_currentLogFileHandle offsetInFile];
         }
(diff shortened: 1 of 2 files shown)


Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -650,9 +650,14 @@ - (instancetype)initWithLogFileManager:(id <DDLogFileManager>)aLogFileManager
 - (void)lt_cleanup {
     NSAssert([self isOnInternalLoggerQueue], @"lt_ methods should be on logger queue.");
 
-    [_currentLogFileHandle synchronizeFile];
-    [_currentLogFileHandle closeFile];
-
+    if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
+        [_currentLogFileHandle synchronizeAndReturnError:nil];
```
lolgear, line 654: Are you sure that we need to hide possible error?
I would like to know if something strange has happened.

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -947,7 +957,12 @@ - (void)lt_maybeRollLogFileDueToSize {
     // We specifically wrote our own getter/setter method to allow us to do this (for performance reasons).
 
     if (_maximumFileSize > 0) {
-        unsigned long long fileSize = [_currentLogFileHandle offsetInFile];
+        unsigned long long fileSize;
+        if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
+            [_currentLogFileHandle getOffset:&fileSize error:nil];
```
lolgear, line 962: Again, if they give us API with errors, why we should ignore it?

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1144,7 +1159,11 @@ - (NSFileHandle *)lt_currentLogFileHandle {
     if (!_currentLogFileHandle) {
         NSString *logFilePath = [[self lt_currentLogFileInfo] filePath];
         _currentLogFileHandle = [NSFileHandle fileHandleForWritingAtPath:logFilePath];
-        [_currentLogFileHandle seekToEndOfFile];
+        if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
+            [_currentLogFileHandle seekToEndReturningOffset:nil error:nil];
```
lolgear, line 1163: Again, errors are our friends, don't ignore them.

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1295,8 +1318,13 @@ - (void)lt_logData:(NSData *)data {
         }
 
         NSFileHandle *handle = [self lt_currentLogFileHandle];
-        [handle seekToEndOfFile];
-        [handle writeData:data];
+        if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
+            [handle seekToEndReturningOffset:nil error:nil];
+            [handle writeData:data error:nil];
```
lolgear, line 1323: Errors...

lolgear:
New API is ok, but I would like to show errors to a developer, of course.

kinarobin:
> New API is ok, but I would like to show errors to a developer, of course.

Emm, It's more appropriate to show errors.

Sources/CocoaLumberjack/DDFileLogger.m
kinarobin: Done!

Sources/CocoaLumberjack/DDFileLogger.m
kinarobin: Yes, we shouldn't ignore it!

Sources/CocoaLumberjack/DDFileLogger.m
kinarobin: Done!

Sources/CocoaLumberjack/DDFileLogger.m
kinarobin: Done!

lolgear:
It is fine for me, but I don't remember if `os_log` is available for current os minimum target. I would like to hear some thoughts about this API from @CocoaLumberjack/collaborators

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -947,7 +971,16 @@ - (void)lt_maybeRollLogFileDueToSize {
     // We specifically wrote our own getter/setter method to allow us to do this (for performance reasons).
 
     if (_maximumFileSize > 0) {
-        unsigned long long fileSize = [_currentLogFileHandle offsetInFile];
+        unsigned long long fileSize;
+        if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
+            __autoreleasing NSError *error;
+            BOOL succeed = [_currentLogFileHandle getOffset:&fileSize error:&error];
+            if (!succeed) {
+                NSLogError(@"DDFileLogger: Failed to get offset: %@", error);
```
ffried, line 979: I wonder if we should abort (return) here. Or at least explicitly set the `fileSize` to 0. Would be interesting what the old `-offsetInFile` returned in case of failure.

ffried: Ok, just checked. `-offsetInFile` raised an exception. So I'm really in favor of adding a `return` here.

ffried (changes requested):
Looks good, thanks! In most cases there are no return values, so except for logged errors there's not much new. NSFileHandle did raise exceptions, but I think it's a good thing that we no longer rely on that and have error handling in our own hands.
It's just one case where we actually retrieve a value where we need to handle the error case a bit better by at least returning now (see inline comment).

ffried:
Also, please add a CHANGELOG entry. 🙂

Sources/CocoaLumberjack/DDFileLogger.m
kinarobin: In case of failure, fileSize is 0 in default.

But `return` is a good idea!

kinarobin:
> Also, please add a CHANGELOG entry. 🙂

Done!

ffried (approved):

Sources/CocoaLumberjack/DDFileLogger.m
kinarobin: `return` Done!

sushichop (commented):
@lolgear 

> It is fine for me, but I don't remember if os_log is available for current os minimum target. I would like to hear some thoughts about this API from @CocoaLumberjack/collaborators

`os_log` is available from macOS 10.12, iOS 10.0, tvOS 10.0, and watchOS 3.0.
But this PR is about new API about NSFileHandle and FileLogger, so this PR is not related to `os_log`.

sushichop (approved):

PR #1184, opened 2020-11-09
Add an assertion to avoid potential deadlock issues for `flushLog`

kinarobin:

The `loggerQueue` is public, once the user call `flushLog` in the `loggerQueue`, deadlock will occur.

...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Using setter to replace kvo for `NSFileLogger` (#1180)
 - Remove unnecessary checks in `DDFileLogger` (#1182)
+- Add an assertion to avoid potential deadlock issues for `flushLog` (#1183)
 
 ## [3.7.0 - Xcode 12 on Oct 2nd, 2020](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.7.0)
 
--- a/Sources/CocoaLumberjack/DDLog.m
+++ b/Sources/CocoaLumberjack/DDLog.m
@@ -521,6 +521,9 @@ static NSUInteger _numProcessors;
 }
 
 - (void)flushLog {
+    NSAssert(!dispatch_get_specific(GlobalLoggingQueueIdentityKey),
+             @"This method shouldn't be run on the logging thread/queue that make flush fast enough");
+    
     dispatch_sync(_loggingQueue, ^{ @autoreleasepool {
         [self lt_flush];
     } });


ffried:
I'm not sure this is really a problem. The `loggerQueue` is public for custom logger implementations, not for arbitrary use.
Given that `flushLogs` needs to be fast, we should probably skip this check. Calling `flushLogs` from the `loggerQueue` is a programmer error.

I could imagine having an assertion to track calls to this methods happening from the `loggerQueue`, so that in DEBUG builds, developers are informed about the mistake (instead of just deadlocking).

kinarobin:
From this code: 
```objc
- (void)queueLogMessage:(DDLogMessage *)logMessage asynchronously:(BOOL)asyncFlag {
    
    dispatch_block_t logBlock = ^{
        dispatch_semaphore_wait(_queueSemaphore, DISPATCH_TIME_FOREVER);
        // We're now sure we won't overflow the queue.
        // It is time to queue our log message.
        @autoreleasepool {
            [self lt_log:logMessage];
        }
    };

    if (asyncFlag) {
        dispatch_async(_loggingQueue, logBlock);
    } else if (dispatch_get_specific(GlobalLoggingQueueIdentityKey)) {
        // We've logged an error message while on the logging queue...
        logBlock();
    } else {
        dispatch_sync(_loggingQueue, logBlock);
    }
}
```

case1
```objc
dispatch_sync(DDLog.loggingQueue, ^{
    DDLogInfo(@"info");
});
```

case2
```objc
dispatch_sync(DDLog.loggingQueue, ^{
    [DDLog flushLog];
});
```


If you think calling `flushLogs` from the `loggerQueue` is a programmer error. Is the error that programmer queueLog in `loggerQueue` ? 

So I think case2 is OK.

kinarobin:
In other words, we should not focus on how programmer use it in a wrong way, but make the program stronger enough.

ffried:
Doing `dispatch_sync(DDLog.loggingQueue, ^{ /*anything*/ })` is something you shouldn't do at all (unless you're writing your own logger - in which case you would still not flush _all_ logs).

The header docs for `DDLog.loggingQueue` states, that it is there for custom loggers. Apart from that, `DDLog.loggingQueue` is under the jurisdiction of `DDLog` and there is no guarantee that arbitrarily dispatching onto it won't lead to deadlocks.

It is true, that this might prevent a potential deadlock. But at the cost of performance, which in this case is quite relevant, given that it is e.g. called from application termination handler - and this has to be _fast_.

The check in `queueLogMessage: asynchronously :` is not as critical, given that it's only relevant for synchronously logged messages (which by default are only error messages).

Long story short: This would optimize a place that needs to be fast and with which we never had any issues so far (AFAIK).

kinarobin:
I get your idea,  how about to set an assertion if programmer use in in a wrong way?

ffried:
Yes, that's what I meant by this:
> I could imagine having an assertion to track calls to this methods happening from the loggerQueue, so that in DEBUG builds, developers are informed about the mistake (instead of just deadlocking).

ffried (approved):

ffried (approved):

Issue #1185, opened 2020-11-10
Log has lost when user manually kill app with `[fileLogger wrapWithBuffer]` 


kinarobin:
### New Issue Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](https://github.com/CocoaLumberjack/CocoaLumberjack#documentation) or [CocoaDocs](cocoadocs.org/docsets/CocoaLumberjack)
* [x] I have searched for a similar issue in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/issues) and found none

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform Name           | e.g. ios 
 Platform Version        | e.g. 13.5.1
 CocoaLumberjack Version | e.g. 3.7.0
 Integration Method      | e.g. cocoapods 
 Xcode Version           | e.g. Xcode 12.1
 Repro rate              | e.g. all the time (100%) 

### Issue Description and Steps

Please fill in the detailed description of the issue (full textual output of any stack trace, compiler error, ...) and the steps to reproduce the issue.

steps:
1. use `[DDLog addLogger:[fileLogger wrapWithBuffer]];`
2. back the app to the background
3. kill the app from background

Then the log had lost.

ffried:
As mentioned in [this comment](https://github.com/CocoaLumberjack/CocoaLumberjack/pull/1186#issuecomment-724625427), this is somewhat intentional. It's the tradeoff for improved speeds. As also mentioned in the comment, there is an easy way to work around this if this is an issue in client apps.

PR #1186, opened 2020-11-10
Flush data when app enter background

kinarobin:

This merge request fixes / refers to the following issues: #1185 

### Pull Request Description

There is a risk in flushing data when the app is about to be terminated. 

`DDFileLogger+Buffering` should not rely heavily on flushing when the app is about to be terminated. Once the `applicationwillterminate:` method is not called, the log data had lost.  
I think we should flush data in background as soon as possible instead of app terminated. 

Because the background mechanism of AppKit and UIKit are different, we can not care about the background of AppKit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Use new API for `NSFileHandle` on supported platforms (#1181)
 - Remove unnecessary checks in `DDFileLogger` (#1182)
 - Add an assertion to avoid potential deadlock issues for `flushLog` (#1183)
-- Flush data when app enter background (#1184)
+- Flush data when app enter background (#1186)
 
 ## [3.7.0 - Xcode 12 on Oct 2nd, 2020](https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.7.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Use new API for `NSFileHandle` on supported platforms (#1181)
 - Remove unnecessary checks in `DDFileLogger` (#1182)
 - Add an assertion to avoid potential deadlock issues for `flushLog` (#1183)
+- Flush data when app enter background (#1184)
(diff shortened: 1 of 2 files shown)


ffried:
FileLogger has two modes:
- unbuffered -> Is slightly slower than buffered but has a very very low risk of loosing messages.
- buffered     -> Is faster, but the tradeoff is the risk of loosing messages when the app terminates.

There are a few suboptimal things here:
- The newly observed notification corresponds to an `UIApplicationDelegate` method that is no longer called if the application uses (window-)scenes. While the docs states, that the notification is still posted, I'm not sure for how long that will be.
- Starting background tasks is logged by iOS and takes away time the app might want for doing other (more important) things.
- Depending on custom logger implementations, this will introduce a serious overhead.

IMHO, this is something we shouldn't do. Again, we (AFAIK) never had any reports of this being a problem. And it's something an app can totally decide for itself. Like if this is a problem in your app, you can always add a `[DDLog flushLogs];` (or `DDLog.flushLogs()` in Swift) in your AppDelegate or SceneDelegate in the corresponding method.

The termination handler we have is really something _every_ application needs. The background handler, however, is not needed by every application and thus is IMHO something we shouldn't ship by default.
I'm not sure if we have a section about this somewhere in the docs, but I'd rather have a sample snippet there somewhere for apps that have this particular problem.

Also, the amount of code this feature would add to CocoaLumberjack versus the amount of code that is needed to add this to an AppDelegate or SceneDelegate (about one line) is IMHO another reason why this should go into client apps instead of the lib.

kinarobin:
Thank you very much for your detailed explanation.  Maybe there is something wrong with my use case. I'm going to close it.

ffried:
@kinarobin If you have this specific problem, please do try to add the `-flushLogs` call to your AppDelegate or SceneDelegate. That should have the same effect and it's totally possible that it is needed in your case! 🙂

kinarobin:
Yes, Thank you!

PR #1209, opened 2021-02-02
Add per-message synchronous logging control for DDLogHandler

ffried:

This merge request fixes / refers to the following issues: n/a

### Pull Request Description

With CocoaLumberjack and CocoaLumberjackSwift it's possible to control synchronous/asynchronous logging on a per-message basis. Our implementation of a SwiftLog backend currently couldn't do this.

With this PR it's possible to do this using a metadata-value on the message. The key for this metadata-value is configurable, but has a default of `"log-synchronous"`. Said default is accessible via `DDLogHandler.defaultSynchronousLoggingMetadataKey`.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -8,3 +8,6 @@ opt_in_rules:
 
 nesting:
   type_level: 4
+
+identifier_name:
+  allowed_symbols: "_"
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -11,4 +11,3 @@ nesting:
 
 identifier_name:
   allowed_symbols: "_"
-  validates_start_with_lowercase: false
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -11,3 +11,4 @@ nesting:
 
 identifier_name:
   allowed_symbols: "_"
+  validates_start_with_lowercase: false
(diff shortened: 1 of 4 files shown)


Sources/CocoaLumberjackSwiftLogBackend/DDLogHandler.swift
```diff
@@ -135,12 +142,14 @@ public struct DDLogHandler: LogHandler {
...
+    @inlinable
     public var logLevel: Logger.Level {
-        get { loggerInfo.logLevel }
+        get { return loggerInfo.logLevel }
         set { loggerInfo.logLevel = newValue }
     }
+    @inlinable
     public var metadata: Logger.Metadata {
-        get { loggerInfo.metadata }
+        get { return loggerInfo.metadata }
```
ffried, line 152: We would have actually had a problem building this module against Swift versions pre 5.2 (or 5.1?) due to those missing returns. 🙈 

Either way, I guess we can soon drop support for those Swift versions and use those nice implicit returns...

.swiftlint.yml
```diff
@@ -8,3 +8,6 @@ opt_in_rules:
 
 nesting:
   type_level: 4
+
+identifier_name:
+  allowed_symbols: "_"
```
ffried, line 13: Beginning underlines are common for `@usableFromInline internal` stuff in Swift. Or generally for adding another layer of access control where performance is the bigger issue. It generally states: "You might see this symbol, but please refrain from using it unless you know what you're doing."

sushichop (approved):
LGTM👍

Issue #1214, opened 2021-02-17
Remove Non-Inclusive Terms


tinder-owenthomas:
### Issue Info

CocoaLumberjack currently has several references to non-inclusive terms.

### Issue Description and Steps

Can we please get an updated version of the SDK which removes all non-inclusive language? We are focusing on removing such language from our company which extends to all third-party code we choose to use.

Specifically terms such as [whitelist](https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=whitelist) (also [here](https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=addToWhitelist), [here](https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=removeFromWhitelist), and [here](https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=isOnWhitelist)), [blacklist](https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=blacklist) (also [here](https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=addtoblacklist), and [here](https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=removefromblacklist)), [master](https://github.com/CocoaLumberjack/CocoaLumberjack/search?q=master), slave, etc. Would love this to cover actual code of course, but also, comments, branch names, repo descriptions on dependency management central repositories, licensing agreements, etc.

ffried:
Thanks for bringing this to our attention!
IMHO this needs a team discussion here (pinging @CocoaLumberjack/collaborators).
<hr>
My two cents:

I think CocoaLumberjack doesn't have too many references. The blacklist/whitelist ones should be rather straightforward to deprecate and replace with blocklist/allowlist (<- my suggestion for the new names).

As for `master`: My personal opinion here is that in the context of `git` renaming `master` is nonsense. While I don't agree with everything, [this blog post](https://felipec.wordpress.com/2020/11/13/git-master/) lists some IMHO important points as to why. But from a more objective perspective: Renaming `master` is non-trivial and breaks many integrations. It's not impossible, but it's needs careful consideration. My personal vote goes against it.

Apart from `git` / branches:
There's one `#pragma mark` containing "master". That one's trivial and might not fit the points described in the blog post.
However, there are also many references to `sqlite_master`, which is not something we can change.

lolgear:
Please, stop this madness about naming conventions.

[...]

PR #1218, opened 2021-03-16
Use inclusive words - denylist / allowlist

sushichop:

This merge request fixes / refers to the following issues: #1214

### Pull Request Description

Use inclusive words(Change `blacklist/whitelist` to `denylist/allowlist`).

### DIFF
--- a/Sources/CocoaLumberjack/include/CocoaLumberjack/DDContextFilterLogFormatter+Deprecated.h
+++ b/Sources/CocoaLumberjack/include/CocoaLumberjack/DDContextFilterLogFormatter+Deprecated.h
@@ -48,7 +48,7 @@ typedef DDContextAllowlistFilterLogFormatter DDContextWhitelistFilterLogFormatte
  *
  *  @param loggingContext the context
  */
-- (void)addToWhitelist:(NSInteger)loggingContext __attribute__((deprecated("Use -addToWhitelist: instead")));
+- (void)addToWhitelist:(NSInteger)loggingContext __attribute__((deprecated("Use -addToAllowlist: instead")));
 
 /**
  *  Remove context from whitelist
--- a/Sources/CocoaLumberjack/include/CocoaLumberjack/DDContextFilterLogFormatter+Deprecated.h
+++ b/Sources/CocoaLumberjack/include/CocoaLumberjack/DDContextFilterLogFormatter+Deprecated.h
@@ -1,0 +1,119 @@
+// Software License Agreement (BSD License)
+//
+// Copyright (c) 2010-2021, Deusty, LLC
+// All rights reserved.
+//
+// Redistribution and use of this software in source and binary forms,
+// with or without modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
+// * Neither the name of Deusty nor the names of its contributors may be used
+//   to endorse or promote products derived from this software without specific
+//   prior written permission of Deusty, LLC.
+
+#import <CocoaLumberjack/DDContextFilterLogFormatter.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * This class provides a log formatter that filters log statements from a logging context not on the whitelist.
+ * @deprecated Use DDContextAllowlistFilterLogFormatter instead.
+ *
+ * A log formatter can be added to any logger to format and/or filter its output.
+ * You can learn more about log formatters here:
+ * Documentation/CustomFormatters.md
+ *
+ * You can learn more about logging context's here:
+ * Documentation/CustomContext.md
+ *
+ * But here's a quick overview / refresher:
+ *
+ * Every log statement has a logging context.
+ * These come from the underlying logging macros defined in DDLog.h.
+ * The default logging context is zero.
+ * You can define multiple logging context's for use in your application.
+ * For example, logically separate parts of your app each have a different logging context.
+ * Also 3rd party frameworks that make use of Lumberjack generally use their own dedicated logging context.
+ **/
+__attribute__((deprecated("Use DDContextAllowlistFilterLogFormatter instead")))
+typedef DDContextAllowlistFilterLogFormatter DDContextWhitelistFilterLogFormatter;
+
+@interface DDContextAllowlistFilterLogFormatter (Deprecated)
+
+/**
+ *  Add a context to the whitelist
+ *  @deprecated Use -addToAllowlist: instead.
+ *
+ *  @param loggingContext the context
+ */
+- (void)addToWhitelist:(NSInteger)loggingContext __attribute__((deprecated("Use -addToWhitelist: instead")));
+
+/**
+ *  Remove context from whitelist
+ *  @deprecated Use -removeFromAllowlist: instead.
+ *
+ *  @param loggingContext the context
+ */
+- (void)removeFromWhitelist:(NSInteger)loggingContext __attribute__((deprecated("Use -removeFromAllowlist: instead")));
+
+/**
+ *  Return the whitelist
+ *  @deprecated Use allowlist instead.
+ */
+@property (nonatomic, readonly, copy) NSArray<NSNumber *> *whitelist __attribute__((deprecated("Use allowlist instead")));
+
+/**
+ *  Check if a context is on the whitelist
+ *  @deprecated Use -isOnAllowlist: instead.
+ *
+ *  @param loggingContext the context
+ */
+- (BOOL)isOnWhitelist:(NSInteger)loggingContext __attribute__((deprecated("Use -isOnAllowlist: instead")));
+
+@end
+
+
+/**
+ * This class provides a log formatter that filters log statements from a logging context on the blacklist.
+ * @deprecated Use DDContextDenylistFilterLogFormatter instead.
+ **/
+__attribute__((deprecated("Use DDContextDenylistFilterLogFormatter instead")))
+typedef DDContextDenylistFilterLogFormatter DDContextBlacklistFilterLogFormatter;
+
+@interface DDContextDenylistFilterLogFormatter (Deprecated)
+
+/**
+ *  Add a context to the blacklist
+ *  @deprecated Use -addToDenylist: instead.
+ *
+ *  @param loggingContext the context
+ */
+- (void)addToBlacklist:(NSInteger)loggingContext __attribute__((deprecated("Use -addToDenylist: instead")));
+
+/**
+ *  Remove context from blacklist
+ *  @deprecated Use -removeFromDenylist: instead.
+ *
+ *  @param loggingContext the context
+ */
+- (void)removeFromBlacklist:(NSInteger)loggingContext __attribute__((deprecated("Use -removeFromDenylist: instead")));
+
+/**
+ *  Return the blacklist
+ *  @deprecated Use denylist instead.
+ */
+@property (readonly, copy) NSArray<NSNumber *> *blacklist __attribute__((deprecated("Use denylist instead")));
+
+/**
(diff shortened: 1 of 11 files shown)


lolgear:
I guess we should change DenyLists, because Deny is offensive to all people.
[Deny](https://en.wikipedia.org/wiki/Deny_(album)) is an album of Ex Pistols band.
It contains a song with a title "Nicotine Addict".
We should change it to ForbidList.

Also "Allow" is not a good term.
[Allow](https://en.wiktionary.org/wiki/allow) has a meaning of granting freedom to someone.
"to allow a servant his liberty".
We should change it to AcceptList, I guess. Or PermitList.

sushichop:
@lolgear 
Thanks! Your explanation made sense to me(but I didn't know the album of Ex Pistols band 😅).
And I prefer `AcceptList` to `PermitList`.

@CocoaLumberjack/collaborators 
Any other opinions? If there are no opinions after waiting for a while, I'll use `forbiddenList/acceptableList`.

Thanks in advance.

ffried (changes requested):
First of all, we need to provided deprecated interfaces here. The worst thing we can do is simply break stuff for existing users.

As for the naming: I think allowList and denyList is what [others](https://inclusivenaming.org/language/word-list/) chose as well, so that's what we should do, too. Again, if we go with something that's not logical, we make a lot of users angry.

lolgear:
@sushichop 
Sorry!
I thought it was obvious and will be treated as a joke :)
( But "Allow" is still something that "master-slave" is ).

lolgear:
@ffried 
Others are illogical.
Because Whitelisting is...

"Whitelisting (also referred to as allow-listing[1]) is the practice of **_explicitly allowing some identified entities access to a particular privilege_**, service, mobility, access or recognition."

Changing a name doesn't changing a meaning.

sushichop (commented):
@ffried 

Thanks for your review.

I added deprecated annotations and unit tests about `DDContextAllowlistFilterLogFormatter/DDContextWhitelistFilterLogFormatter/DDContextDenylistFilterLogFormatter/DDContextBlacklistFilterLogFormatter`.

Sources/CocoaLumberjack/include/CocoaLumberjack/DDContextFilterLogFormatter.h
```diff
@@ -43,6 +99,7 @@ NS_ASSUME_NONNULL_BEGIN
  * For example, logically separate parts of your app each have a different logging context.
  * Also 3rd party frameworks that make use of Lumberjack generally use their own dedicated logging context.
  **/
+__attribute__((deprecated("Use DDContextAllowlistFilterLogFormatter instead")))
 @interface DDContextWhitelistFilterLogFormatter : NSObject <DDLogFormatter>
```
ffried, line 103: I wonder if we can solve this with a `typedef`? Maybe I've been doing Swift for too long now, but in Swift I'd simply create a `typealias OldProtocol = NewProtocol` and deprecate the `typealias`.

ffried: If that works, we wouldn't need two implementations...

ffried (approved):
Looks good in general! Just wondering if we could save some code using `typedef`s...

Sources/CocoaLumberjack/include/CocoaLumberjack/DDContextFilterLogFormatter.h
lolgear: It works, of course, consider `typealias TimeInterval = Double` in Swift.
I guess that @sushichop chose second approach via subclassing/cloning.

Second approach is good for distribution, of course.
But we could achieve the same by adding additional files with "non-inclusive" declarations/type aliases.

```objective-c
/// Non-Inclusive file. 
/// We should add notes about its non-inclusiveness at the header.

/// deprecated annotation
typealias OldSomething = NewSomething

/// deprecated annotation
@interface OldSomething
- (void) someOldMethod; //  { [self someNewMethod]; } in implementation.
@end
```

sushichop: Thanks all,
AFAIK, we can use typealias for type in Objective-C(I think typealias should come from C++ feature.). On the other hand, I often see people use typealias only about blocks type in Objective-C. So I prefer subclassing and chose it as @lolgear said. 

Thanks.

ffried: It's `typedef` in Objective-C and `typealias` in Swift.

We could combine the two: Have a deprecated `typedef` for the old class name (pointing to the new one) and have an extension interface for the old methods (calling the new ones).

A subclass wouldn't be necessary then, and we can put all the "non-inclusive" old stuff into one file (well two, actually, one header one implementation).

sushichop: I'm a bit confused, I might have worded it wrong😅

For example, I think we can use DDContextDenylistFilterLogFormatter itself as a typealias for DContextBlacklistFilterLogFormatter because it is a type(not method and protocol). 
However, DDContextBlacklistFilterLogFormatter has addToBlacklist: method.

I thought I couldn't use DDContextDenylistFilterLogFormatter:addToDenlylist as a typealias for 
DDContextBlacklistFilterLogFormatter:addToBlacklist. We can only call DDContextDenylistFilterLogFormatter:addToDenlylist inside DDContextBlacklistFilterLogFormatter:addToBlacklist.

And we need still provide DDContextBlacklistFilterLogFormatter:addToBlacklist method. It is deprecated method.

Is this wrong?

lolgear: 1. typedef BlacklistFormatter = DenylistFormatter
2. All methods with "Blacklist" should have implementation as follows: blackListMethod { denyListMethod }
3. Deprecate typedef 
4. Deprecate all blacklist methods.

Final steps

1. Put everything in CocoaLumberjack+NonInclusive.h(m)
2. Change Podspec/Package.swift if needed.

lolgear: Actually, you could use "method aliasing" via NSProxy, but don't, please. It is easy to start and hard to finish. Something like NSString(cmd.selector).replace("BlackList", "DenyList").toSelector. ( Sorry, hard to write from phone ).

sushichop: > We could combine the two: Have a deprecated `typedef` for the old class name (pointing to the new one) and have an extension interface for the old methods (calling the new ones).

Thanks. And I overlooked this.
I'll do it.

ffried: @lolgear correctly described it. `typedef` the classes and add the old (deprecated) methods in extensions of those classes... I'd put it in `DDContextFilterLogFormatter+Deprecated.h` or `DDContextFilterLogFormatter+NonInclusive.h`, though. The Package.swift doesn't need to be adjusted (as long as the header file is placed alongside the other public headers).

sushichop: Many thanks!🙂

lolgear: @ffried @sushichop 
Last question.

Will it "required" to distribute "different" packages for "all" and for "non-inclusive"?
I mean "should" we create TWO packages ( two targets for cocoa pods ), one of them will include "non-inclusive" header, and another will not include this header?

Tests/CocoaLumberjackTests/DDContextFilterLogFormatterTests.m
```diff
@@ -0,0 +1,176 @@
...
+    self.filterLogFormatter = [[DDContextBlacklistFilterLogFormatter alloc] init];
+}
+
+- (void)tearDown {
+    self.filterLogFormatter = nil;
+    [super tearDown];
+}
+
+- (void)testDDContextBlacklistFilterLogFormatterTests {
+    XCTAssertEqualObjects([self.filterLogFormatter blacklist], @[]);
```
lolgear, line 160: Move this method to `DDContextDenylistFilterLogFormatterTests` and add deprecated attribute.
Also add comment "for compatibility".

Tests/CocoaLumberjackTests/DDContextFilterLogFormatterTests.m
```diff
@@ -0,0 +1,176 @@
...
+    [super setUp];
+    self.filterLogFormatter = [[DDContextWhitelistFilterLogFormatter alloc] init];
+}
+
+- (void)tearDown {
+    self.filterLogFormatter = nil;
+    [super tearDown];
+}
+
+- (void)testWhitelistFilterLogFormatter {
```
lolgear, line 84: Also here, we could have one test suite with two methods.
One method for new "allow" methods.
Another for old "whitelist" methods.

Sources/CocoaLumberjack/include/CocoaLumberjack/DDContextFilterLogFormatter.h
ffried: IMHO this would only create unnecessary overhead and again break existing integrations. Since it's broken, existing users will have to fix it - if they want it or not. If it's deprecated, they're nudged into the right direction, but have no immediate todo.

As mentioned in the original issue, I'm all for such changes, as long as we don't break stuff for existing users - giving them a reason to be pissed about this topic, which would counteract what the movement wants to achieve.

sushichop:
I have updated PR and resolved conflict.

I used typedef and category for deprecated classes and methods. 
As a result, when we remove the deprecated ones in the future, I think we should only need to delete deprecated {h,m} files and remove `#import <CocoaLumberjack/DDContextFilterLogFormatter+Deprecated.h>` from `CocoaLuimberjack.h` .

Thanks.

Sources/CocoaLumberjack/include/CocoaLumberjack/DDContextFilterLogFormatter+Deprecated.h
```diff
@@ -0,0 +1,119 @@
...
+
+@interface DDContextAllowlistFilterLogFormatter (Deprecated)
+
+/**
+ *  Add a context to the whitelist
+ *  @deprecated Use -addToAllowlist: instead.
+ *
+ *  @param loggingContext the context
+ */
+- (void)addToWhitelist:(NSInteger)loggingContext __attribute__((deprecated("Use -addToWhitelist: instead")));
```
ffried, line 51: Message is incorrect: Should be `"Use -addToAllowlist: instead"`

ffried (changes requested):
Just a small mistake left 🙂

Sources/CocoaLumberjack/include/CocoaLumberjack/DDContextFilterLogFormatter+Deprecated.h
sushichop: Thanks! Fixed it.

ffried (approved):
Looks good now, thanks!

sushichop:
Thanks is mine🙂

bpoplauschi (approved):
Looks good, thanks @sushichop for doing this.

HermanKayy:
@sushichop @ffried Thanks for doing this! It looks like in `DDContextFilterLogFormatter+Deprecated.h` and `DDContextFilterLogFormatter+Deprecated.m` still have a few places leftover that's referencing `blacklist`

PR #1234, opened 2021-06-30
Fix "DDFileLogger: Failed to get offset" when setting maximumFileSize

guojiubo:

### Pull Request Description

DDFileLogger will fail to get the offset of the log file when `_currentLogFileHandle` isn't initialized, and the log error message "DDFileLogger: Failed to get offset: (null)" will appear.

This happens when the `maximumFileSize` of a DDFileLogger instance is set immediately after the instance is created, for example:
```objc
fileLogger = [[DDFileLogger alloc] init];
    
fileLogger.maximumFileSize = 1024 * 1;  //  1 KB
fileLogger.rollingFrequency = 60;       // 60 Seconds
```
The code is copied from `Demos/RollingTestMac`, and I believe this is a common usage. The issue can also be reproduced in the demo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Public
 
-- _REPLACE ME_
+- Fix "DDFileLogger: Failed to get offset" when setting maximumFileSize (#1234)
 
 ### Internal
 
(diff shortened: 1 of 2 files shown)


ffried (approved):

ffried:
Thanks for fixing this!

PR #1240, opened 2021-07-18
Fixed false positive error reports.

jcbertin:
There is a problem when we call new `NSFileHandle` APIs that return errors: if the `_currentLogFileHandle` is `nil`, it will return both `FALSE` for result and `nil` for the error. This will generate false error reports with `NSLogError()`.
There is two ways for fixing this:

1. Check both result set to `FALSE` and error being not `nil` before calling `NSLogError()`.
2. Don't call the API if `_currentLogFileHandle` is `nil` at first place!

I preferred the second solution because it will save CPU cycles.

I did also not include PR #1234 on purpose because I think my solution is better. Let me explain:

1. The way it is fixed in PR #1234, breaks the comment just above that says "Keep it FAST"!
2. If `_currentLogFileHandle` is `nil` when `-lt_maybeRollLogFileDueToSize` is called, it is either because `-lt_logData:` was not called or because `-lt_currentLogFileHandle` failed. In both cases, it means that the file size was not increased and don't need to be rolled. One side effect of this is that the file will be not rolled if `maximumFileSize` is set to something lower than previous `maximumFileSize` and actual file size is greater than the `maximumFileSize` being set, but it is the way it worked before using these new `NSFileHandle` APIs.

### DIFF
--- a/Sources/CocoaLumberjack/DDFileLogger.m
+++ b/Sources/CocoaLumberjack/DDFileLogger.m
@@ -637,7 +637,11 @@ NSTimeInterval     const kDDRollingLeeway              = 1.0;              // 1s
                 NSLogError(@"DDFileLogger: Failed to close file: %@", error);
             }
         } else {
-            [_currentLogFileHandle synchronizeFile];
+            @try {
+                [_currentLogFileHandle synchronizeFile];
+            } @catch (NSException *exception) {
+                NSLogError(@"DDFileLogger: Failed to synchronize file: %@", exception);
+            }
             [_currentLogFileHandle closeFile];
         }
         _currentLogFileHandle = nil;
@@ -887,7 +891,11 @@ NSTimeInterval     const kDDRollingLeeway              = 1.0;              // 1s
             NSLogError(@"DDFileLogger: Failed to close file: %@", error);
         }
     } else {
-        [_currentLogFileHandle synchronizeFile];
+        @try {
+            [_currentLogFileHandle synchronizeFile];
+        } @catch (NSException *exception) {
+            NSLogError(@"DDFileLogger: Failed to synchronize file: %@", exception);
+        }
         [_currentLogFileHandle closeFile];
     }
     _currentLogFileHandle = nil;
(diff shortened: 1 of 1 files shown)


ffried (approved):
Thanks for the improvements!

jcbertin:
If we want to keep PR #1234 behavior where setting maximumSize to a lower value than the previous one will immediately roll the log file, we may apply this patch:
```
diff --git a/Sources/CocoaLumberjack/DDFileLogger.m b/Sources/CocoaLumberjack/DDFileLogger.m
index dc338a0..2c26a99 100644
--- a/Sources/CocoaLumberjack/DDFileLogger.m
+++ b/Sources/CocoaLumberjack/DDFileLogger.m
@@ -705,7 +705,9 @@ NSTimeInterval     const kDDRollingLeeway              = 1.0;              // 1s
     dispatch_block_t block = ^{
         @autoreleasepool {
             self->_maximumFileSize = newMaximumFileSize;
-            [self lt_maybeRollLogFileDueToSize];
+            if (_currentLogFileHandle != nil || [self lt_currentLogFileHandle] != nil) {
+                [self lt_maybeRollLogFileDueToSize];
+            }
         }
     };
```

Are you agree?

ffried:
@jcbertin Sounds good! 👍

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -698,7 +705,9 @@ - (void)setMaximumFileSize:(unsigned long long)newMaximumFileSize {
     dispatch_block_t block = ^{
         @autoreleasepool {
             self->_maximumFileSize = newMaximumFileSize;
-            [self lt_maybeRollLogFileDueToSize];
+            if (_currentLogFileHandle != nil || [self lt_currentLogFileHandle] != nil) {
```
sushichop, line 708: Use `self` explicitly in block to fix `implicit-retain-self warning`.
```objc
self->_rollingFrequency
```

sushichop (commented):
@jcbertin 
Thanks for doing this! 
`pod lib lint` did not pass validation. I left a comment to fix it.

Thanks in advance.

sushichop (approved):

Issue #1243, opened 2021-07-22
iOS Version Issue with DDOSLogger


LowAmmo:
### New Issue Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](https://github.com/CocoaLumberjack/CocoaLumberjack#documentation) or [CocoaDocs](cocoadocs.org/docsets/CocoaLumberjack)
* [x] I have searched for a similar issue in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/issues) and found none

### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform Name           | iOS
 Platform Version        | e.g. 10.0+
 CocoaLumberjack Version | e.g. 3.7.2
 Integration Method      | cocoapods
 Xcode Version           | e.g. Xcode 13 beta 3
 Repro rate              | 100%
 Demo project link       | (just build with Xcode 13)

### Issue Description and Steps

Just seeing a build warning about DDOSLogger only being available in iOS 10.0+, but the deployment target is set to 9.0.

Which is fine, except for the `DDOSLogger *sharedInstance` in DDOSLogger.m - https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/Sources/CocoaLumberjack/DDOSLogger.m#L51

Full error from the build:

```
/Users/.../Pods/CocoaLumberjack/Sources/CocoaLumberjack/DDOSLogger.m:51:8: warning: 'DDOSLogger' is only available on iOS 10.0 or newer [-Wunguarded-availability]
static DDOSLogger *sharedInstance;
       ^
In file included from /Users/.../Pods/CocoaLumberjack/Sources/CocoaLumberjack/DDOSLogger.m:18:
/Users/.../Pods/CocoaLumberjack/Sources/CocoaLumberjack/include/CocoaLumberjack/DDOSLogger.h:31:12: note: 'DDOSLogger' has been marked as being introduced in iOS 10.0 here, but the deployment target is iOS 9.0.0
@interface DDOSLogger : DDAbstractLogger <DDLogger>
           ^
/Users/.../Pods/CocoaLumberjack/Sources/CocoaLumberjack/DDOSLogger.m:51:20: note: annotate 'sharedInstance' with an availability attribute to silence this warning
static DDOSLogger *sharedInstance;
[...]

PR #1248, opened 2021-08-11
Follow-up to add annotations to `DDOSLogger`

sushichop:

This merge request fixes / refers to the following issues: #1243
Since `sharedInstance` is a  class property, it seems that we need to add the annotation to `DDOSLogger.m` , not `DDOSLogger.h`.

### DIFF
--- a/Sources/CocoaLumberjack/DDOSLogger.m
+++ b/Sources/CocoaLumberjack/DDOSLogger.m
@@ -48,6 +48,7 @@
     return self;
 }
 
+API_AVAILABLE(macos(10.12), ios(10.0), watchos(3.0), tvos(10.0))
 static DDOSLogger *sharedInstance;
 
 - (instancetype)init {


ffried:
@sushichop Are we sure this isn't a temporary hiccup with the beta of Xcode 13? It feels weird having to annotate the implementation file 🤔

sushichop:
@ffried 
Good point! Indeed, Xcode 12.5 doesn't seem to show this warning... I just confirmed it. 
(FYI. It still shows up in Xcode 13.0 beta5.)

Shall we wait and see until Xcode 13.5 GM?

ffried:
@sushichop Yea, I'd simply keep the PR around in case it's still necessary once Xcode 13 goes into GM state.
Maybe a feedback report would be a good idea as well - possibly even referencing this issue. 😉

LowAmmo:
Looking at the issue... It seems to be a legit problem - just Xcode 13 seems to be better about static analysis to detect these types of possible problems.

LowAmmo (approved):

ffried:
@LowAmmo I'm still not convinced that it's a real problem. The static sharedInstance is nested inside a class that's correctly annotated. Thus the sharedInstance storage isn't reachable on older platforms since all access paths are sufficiently annotated. Or am I missing something here?

LowAmmo:
Yes, I would agree it's not a "real problem" - it's not going to cause a crash, etc, because it wouldn't get used outside of the DDOSLogger.m anyway.

But, it is a warning, and we  try to get our projects down to 0 warnings - to help ensure that when a more legit/dangerous warning comes up, our developers actually notice it.

(Admittedly, it might be a little OCD... I just really like to build and see no warnings! 😄 )

LowAmmo:
I mean...this is a focused audience of developers worried about logging...

I guess it's safe to assume I'm not the only one that's a little bit OCD on these things... 😉

ffried:
@sushichop Is this still an issue with the RC of Xcode 13?

sushichop:
@ffried Yes, this is still an issue with Xcode 13.0 RC (13A233) unfortunately...

ffried (approved):

ffried:
Well then, let's get this in...

PR #1272, opened 2021-12-23
Add consistent newline to file endings

### LINKED ISSUE #1034: Add missing newlines to the end of some files.
Some files in the repository are missing newlines at the very end, as pointed out by @nrbrook in #1026

`Classes/CocoaLumberjack.h` was specifically mentioned, but there could be other files that are missing this as well.

fdenzer:

* [ ] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [ ] I have read the [Documentation](http://cocoadocs.org/docsets/CocoaLumberjack/)

### Pull Request Description

Fixes #1034 

Via running:
`for x in $(find . -type f); do if [ -n "$(tail -c 1 <"$x")" ]; then echo >>"$x"; fi; done`

I am not sure, if every png needs to end w/ newline, though.


ffried:
Thanks for taking this on!

I think we need to exclude a few files here:

- binary files don't need to be updated (could even corrupt them):
  - with below's exceptions, this is just png files in this case 
- tool-generated files don't need to be updated:
  - everything inside .idea
  - everything inside a *.xcodeproj
  - everything inside a *.xcassets
  - everything inside a *.xcdatamodeld

ffried (changes requested):
Thanks for taking this on!

I think we need to exclude a few files here:

- binary files don't need to be updated (could even corrupt them):
  - with below's exceptions, this is just png files in this case
- tool-generated files don't need to be updated:
  - everything inside .idea
  - everything inside a *.xcodeproj
  - everything inside a *.xcassets
  - everything inside a *.xcdatamodeld

ffried (approved):

PR #1274, opened 2021-12-27
Fix for "DDFileLogger: Failed to synchronize file: nil"

fujisoft:
[_currentLogFileHandle synchronizeAndReturnError:&error] can return NO without error.
DDFileLogger will put "DDFileLogger: Failed to synchronize file: nil" to console.
This patch prevents the problem.

### New Pull Request Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [x] I have read the [Documentation](http://cocoadocs.org/docsets/CocoaLumberjack/)
* [x] I have searched for a similar pull request in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/pulls) and found none

* [x] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)
* [x] I have added the required tests to prove the fix/feature I am adding
* [x] I have updated the documentation (if necessary)
* [x] I have run the tests and they pass
* [ ] I have run the lint and it passes (`pod lib lint`)

This merge request fixes / refers to the following issues: ...

### Pull Request Description

...

PR #1283, opened 2022-02-09
Revert pr 1274

jcbertin:

### Pull Request Description

Please revert PR #1274 because it was not very well discussed. Can we reproduce that "DDFileLogger: Failed to synchronize file: nil" log? Is it occurring in the "@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)" branch of the code?

I prefer to have a log with nil error showing that something went wrong rather than silently ignoring the failed status returned by -synchronizeAndReturnError: method.
Moreover the PR broke -lt_maybeRollLogFileDueToSize: method if the returned status of -getOffset:error: method is false but error is nil: fileSize will be undefined in that case.

Lastly, I have fixed -createNewLogFileWithError: method since error parameter is nullable.

Regards

### DIFF
--- a/Sources/CocoaLumberjack/DDFileLogger.m
+++ b/Sources/CocoaLumberjack/DDFileLogger.m
@@ -220,11 +220,11 @@ unsigned long long const kDDDefaultSecForOneMinute        = 60;
         for (NSUInteger i = firstIndexToDelete; i < sortedLogFileInfos.count; i++) {
             DDLogFileInfo *logFileInfo = sortedLogFileInfos[i];
 
-            __autoreleasing NSError *error;
+            NSError *error = nil;
             BOOL success = [[NSFileManager defaultManager] removeItemAtPath:logFileInfo.filePath error:&error];
             if (success) {
                 NSLogInfo(@"DDLogFileManagerDefault: Deleting file: %@", logFileInfo.fileName);
-            } else if (error) {
+            } else {
                 NSLogError(@"DDLogFileManagerDefault: Error deleting file %@", error);
             }
         }
@@ -258,15 +258,16 @@ unsigned long long const kDDDefaultSecForOneMinute        = 60;
 - (NSString *)logsDirectory {
     // We could do this check once, during initialization, and not bother again.
     // But this way the code continues to work if the directory gets deleted while the code is running.
+
     NSAssert(_logsDirectory.length > 0, @"Directory must be set.");
 
-    __autoreleasing NSError *error;
+    NSError *err = nil;
     BOOL success = [[NSFileManager defaultManager] createDirectoryAtPath:_logsDirectory
                                              withIntermediateDirectories:YES
                                                               attributes:nil
-                                                                   error:&error];
-    if (!success && error) {
-        NSLogError(@"DDFileLogManagerDefault: Error creating logsDirectory: %@", error);
+                                                                   error:&err];
+    if (success == NO) {
+        NSLogError(@"DDFileLogManagerDefault: Error creating logsDirectory: %@", err);
     }
 
     return _logsDirectory;
@@ -707,12 +708,11 @@ unsigned long long const kDDDefaultSecForOneMinute        = 60;
         if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
             __autoreleasing NSError *error;
             BOOL synchronized = [_currentLogFileHandle synchronizeAndReturnError:&error];
-            if (!synchronized && error) {
+            if (!synchronized) {
                 NSLogError(@"DDFileLogger: Failed to synchronize file: %@", error);
-                error = nil;
             }
             BOOL closed = [_currentLogFileHandle closeAndReturnError:&error];
-            if (!closed && error) {
+            if (!closed) {
                 NSLogError(@"DDFileLogger: Failed to close file: %@", error);
             }
         } else {
@@ -725,7 +725,7 @@ unsigned long long const kDDDefaultSecForOneMinute        = 60;
         }
         _currentLogFileHandle = nil;
     }
-
+        
     if (_currentLogFileVnode) {
         dispatch_source_cancel(_currentLogFileVnode);
         _currentLogFileVnode = NULL;
@@ -960,16 +960,15 @@ unsigned long long const kDDDefaultSecForOneMinute        = 60;
     if (_currentLogFileHandle == nil) {
         return;
     }
-
+    
     if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
         __autoreleasing NSError *error;
         BOOL synchronized = [_currentLogFileHandle synchronizeAndReturnError:&error];
-        if (!synchronized && error) {
+        if (!synchronized) {
             NSLogError(@"DDFileLogger: Failed to synchronize file: %@", error);
-            error = nil;
         }
         BOOL closed = [_currentLogFileHandle closeAndReturnError:&error];
-        if (!closed && error) {
+        if (!closed) {
             NSLogError(@"DDFileLogger: Failed to close file: %@", error);
         }
     } else {
@@ -1042,8 +1041,8 @@ unsigned long long const kDDDefaultSecForOneMinute        = 60;
         unsigned long long fileSize;
         if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
             __autoreleasing NSError *error;
-            BOOL succeeded = [_currentLogFileHandle getOffset:&fileSize error:&error];
-            if (!succeeded && error) {
+            BOOL succeed = [_currentLogFileHandle getOffset:&fileSize error:&error];
+            if (!succeed) {
                 NSLogError(@"DDFileLogger: Failed to get offset: %@", error);
                 return;
             }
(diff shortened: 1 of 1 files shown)


Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -220,11 +220,11 @@ - (void)deleteOldLogFiles {
         for (NSUInteger i = firstIndexToDelete; i < sortedLogFileInfos.count; i++) {
             DDLogFileInfo *logFileInfo = sortedLogFileInfos[i];
 
-            __autoreleasing NSError *error;
+            __autoreleasing NSError *error = nil;
             BOOL success = [[NSFileManager defaultManager] removeItemAtPath:logFileInfo.filePath error:&error];
             if (success) {
                 NSLogInfo(@"DDLogFileManagerDefault: Deleting file: %@", logFileInfo.fileName);
-            } else if (error) {
```
ffried, line 227: IMHO this wasn't entirely wrong. Since we don't really handle the error, it seems safer to also check it before logging it.

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -439,38 +440,41 @@ - (NSString *)createNewLogFileWithError:(NSError *__autoreleasing  _Nullable *)e
 
     NSString *fileName = [self newLogFileName];
     NSString *logsDirectory = [self logsDirectory];
-    NSData *fileHeader = [self logFileHeaderData];
-    if (fileHeader == nil) {
-        fileHeader = [NSData new];
-    }
+    NSData *fileHeader = [self logFileHeaderData] ?: [NSData new];
 
+    NSString *baseName = nil;
```
ffried, line 445: These (`baseName` and `extension`) can be declared inline below.

ffried: (Inside the `if (attempt > 1)`)

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -725,7 +724,7 @@ - (void)lt_cleanup {
         }
         _currentLogFileHandle = nil;
     }
-
+
```
ffried, line 727: NitPick: Superfluous whitespace.

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -705,14 +705,13 @@ - (void)lt_cleanup {
...
-            BOOL synchronized = [_currentLogFileHandle synchronizeAndReturnError:&error];
-            if (!synchronized && error) {
+            __autoreleasing NSError *error = nil;
+            BOOL success = [_currentLogFileHandle synchronizeAndReturnError:&error];
+            if (!success) {
                 NSLogError(@"DDFileLogger: Failed to synchronize file: %@", error);
-                error = nil;
             }
-            BOOL closed = [_currentLogFileHandle closeAndReturnError:&error];
-            if (!closed && error) {
```
ffried, line 715: Again, the error check isn't entirely wrong.

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -705,14 +705,13 @@ - (void)lt_cleanup {
 
     if (_currentLogFileHandle != nil) {
         if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
-            __autoreleasing NSError *error;
-            BOOL synchronized = [_currentLogFileHandle synchronizeAndReturnError:&error];
-            if (!synchronized && error) {
```
ffried, line 710: Same here: Checking the error isn't completely wrong

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -705,14 +705,13 @@ - (void)lt_cleanup {
...
+            __autoreleasing NSError *error = nil;
+            BOOL success = [_currentLogFileHandle synchronizeAndReturnError:&error];
+            if (!success) {
                 NSLogError(@"DDFileLogger: Failed to synchronize file: %@", error);
-                error = nil;
```
ffried, line 712: The error should still be reset, to make sure we don't get false positives.

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -960,16 +959,15 @@ - (void)lt_rollLogFileNow {
     if (_currentLogFileHandle == nil) {
         return;
     }
-
+    
     if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
-        __autoreleasing NSError *error;
-        BOOL synchronized = [_currentLogFileHandle synchronizeAndReturnError:&error];
-        if (!synchronized && error) {
```
ffried, line 967: Dito

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -960,16 +959,15 @@ - (void)lt_rollLogFileNow {
...
+        __autoreleasing NSError *error = nil;
+        BOOL success = [_currentLogFileHandle synchronizeAndReturnError:&error];
+        if (!success) {
             NSLogError(@"DDFileLogger: Failed to synchronize file: %@", error);
-            error = nil;
         }
-        BOOL closed = [_currentLogFileHandle closeAndReturnError:&error];
-        if (!closed && error) {
```
ffried, line 972: Dito

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1041,9 +1039,9 @@ - (void)lt_maybeRollLogFileDueToSize {
     if (_currentLogFileHandle != nil && _maximumFileSize > 0) {
         unsigned long long fileSize;
         if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
-            __autoreleasing NSError *error;
-            BOOL succeeded = [_currentLogFileHandle getOffset:&fileSize error:&error];
-            if (!succeeded && error) {
```
ffried, line 1046: Same here.

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1246,9 +1244,9 @@ - (NSFileHandle *)lt_currentLogFileHandle {
         _currentLogFileHandle = [NSFileHandle fileHandleForWritingAtPath:logFilePath];
         if (_currentLogFileHandle != nil) {
             if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
-                __autoreleasing NSError *error;
-                BOOL succeeded = [_currentLogFileHandle seekToEndReturningOffset:nil error:&error];
-                if (!succeeded && error) {
```
ffried, line 1251: Same

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1321,12 +1319,12 @@ - (void)flush {
 
 - (void)lt_flush {
     NSAssert([self isOnInternalLoggerQueue], @"flush should only be executed on internal queue.");
-
+    
     if (_currentLogFileHandle != nil) {
         if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
-            __autoreleasing NSError *error;
-            BOOL synchronized = [_currentLogFileHandle synchronizeAndReturnError:&error];
-            if (!synchronized && error) {
```
ffried, line 1329: And here.

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1417,14 +1415,13 @@ - (void)lt_logData:(NSData *)data {
 
         NSFileHandle *handle = [self lt_currentLogFileHandle];
         if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
-            __autoreleasing NSError *error;
-            BOOL sought = [handle seekToEndReturningOffset:nil error:&error];
-            if (!sought && error) {
```
ffried, line 1422: And here.

ffried (changes requested):
I'm not sure that reverting this will do much good. Checking the errors isn't doing much harm except preventing useless logs.

What problem is this causing you?
Would it be ok to separate the changes from the revert? There are some good cleanup changes here (as well as checking the ptr for null).

ffried:
@jcbertin To answer one of your question: We've seen these logs as well. I'm assuming that these methods might sometimes return false even though nothing went wrong. However, I didn't check how they're annotated for Swift throws conversion. It might be, that Swift simply checks the error pointer for failure and ignores the return value. Didn't have the time so far.

Another case for some of these logs happening was, when the file no longer existed for some reason. Those logs weren't a big deal for us, but did indeed also raise questions in the team whether something was wrong.

jcbertin:
I can assure you that returned boolean is more significant than the error i.e. if a method returns `false`, it means a real failure even if error is `nil`. So removing logs if return value is `false` and error is `nil` will mask real failures to the user which I want to avoid.

Sources/CocoaLumberjack/DDFileLogger.m
jcbertin: No because I want to lazily cache the result of `[fileName stringByDeletingPathExtension]` and `[fileName pathExtension]`.

ffried: Good point.

ffried:
Yea, seems to be the case in the two methods I checked now. Both have no `swift_error` annotation - thus get the default behavior where the result is checked and the error force-unwrapped.

ffried (approved):

PR #1284, opened 2022-02-09
Fixed nullability conflict or file parameter for DDLogMessage creation.

jcbertin:

### Pull Request Description

In PR #1280, `@(file)` could now return nil value. I fixed the nullability conflict by setting file parameter to empty string in that case. Maybe we could use another constant string for that like `@"unknown"`?

### DIFF
--- a/Sources/CocoaLumberjack/DDLog.m
+++ b/Sources/CocoaLumberjack/DDLog.m
@@ -989,8 +989,9 @@ NSString * __nullable DDExtractFileNameWithoutExtension(const char *filePath, BO
                             tag:(id)tag
                         options:(DDLogMessageOptions)options
                       timestamp:(NSDate *)timestamp {
-    NSParameterAssert(message && file);
-    
+    NSParameterAssert(message);
+    NSParameterAssert(file);
+
     if ((self = [super init])) {
         BOOL copyMessage = (options & DDLogMessageDontCopyMessage) == 0;
         _message      = copyMessage ? [message copy] : message;
--- a/Sources/CocoaLumberjack/DDLog.m
+++ b/Sources/CocoaLumberjack/DDLog.m
@@ -462,11 +462,13 @@ static NSUInteger _numProcessors;
    function:(const char *)function
        line:(NSUInteger)line
         tag:(id)tag {
+    NSString* fileString = @(file);
+    NSAssert(fileString, @"Invalid file parameter");
     DDLogMessage *logMessage = [[DDLogMessage alloc] initWithMessage:message
                                                                level:level
                                                                 flag:flag
                                                              context:context
-                                                                file:@(file) ?: @""
+                                                                file:fileString
                                                             function:@(function)
                                                                 line:line
                                                                  tag:tag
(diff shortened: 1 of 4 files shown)


ffried:
Do you know what `[NSString stringWithFormat:@"%s", file]` would have returned in this case? Best case would be to restore that behavior.

jcbertin:
`[NSString stringWithFormat:@"%s", file]` should have returned `(null)` string.

ffried:
I wonder if we should just make the parameter optional - or on the other hand, if we can enforce it being non-null.
Especially since `file` should always be there. `function` might not, but it's hard to have source code, that isn't in some kind of file 🙂

jcbertin:
Of course, `file` is a non null parameter but it is just indicative and you can always initialize a message with `file` set to `nil`. For robustness, we must be prepared for that and handle it correctly.

Sources/CocoaLumberjack/DDLog.m
```diff
@@ -462,11 +462,13 @@ - (void)log:(BOOL)asynchronous
    function:(const char *)function
        line:(NSUInteger)line
         tag:(id)tag {
+    NSString* fileString = @(file);
+    NSAssert(fileString, @"Invalid file parameter");
```
jcbertin, line 466: Added this assertion to catch both `file == nil` and `file` is invalid UTF8 string.

ffried: Isn't `initWithMessage:` a more logical place for that assertion? Otherwise it's still possible to create an invalid `DDLogMessage` using its initializer directly.

sushichop (commented):
Thanks all.
Hmmm... IMO, it might be better to simply revert #1280 partially, in other words, use `[NSString stringWithFormat:@"%s", file]` again.
My understanding and thoughts are as follows.

- `[NSString stringWithFormat:@"%s", file]` does not return `nil`, but may return `(null)` string.
- `@(file)` is a syntax sugar of `[NSString stringWithUTF8String]` in this case(I actually confirmed this using Xcode) and it may return `nil`.
 
For now, I think it may be more complicated to bridge the gap the differences above than simply revert #1280 partially...
Any thoughts?

jcbertin:
@sushichop I think it is relevant to keep #1280:

1. `[NSString stringWithUTF8String]` is faster than `[NSString stringWithFormat:]`
2. Adding parameters validation in `[DDLogMessage -initWithMessage:]` is a good point for robustness of the code

sushichop:
> 1. `[NSString stringWithUTF8String]` is faster than `[NSString stringWithFormat:]`

I see🙂 I didn't know that.

Sources/CocoaLumberjack/DDLog.m
```diff
@@ -985,6 +989,8 @@ - (instancetype)initWithMessage:(NSString *)message
                             tag:(id)tag
                         options:(DDLogMessageOptions)options
                       timestamp:(NSDate *)timestamp {
+    NSParameterAssert(message && file);
```
ffried, line 992: Please use two `NSParameterAssert`s for the two parameters cause the description for a failed assert mentions the parameter name.

jcbertin: Committed.

ffried (approved):

PR #1285, opened 2022-02-09
Fixed -isEqual: and -hash methods for DDLogMessage class.

jcbertin:

### Pull Request Description

This PR will fix -isEqual: and -hash methods for DDLogMessage class:
1. `[super isEqual:other]` always returns `FALSE` since `[NSObject isEqual:]` compares pointers only.
2. `[otherMsg->_function isEqualToString:_function]` always returns `FALSE` if `otherMsg->_function` is `nil`. Added a test for that.
3. Removed the test for `_fileName` since it is a derivative of `_file`.
4. Removed the test for `_options` since it is not relevant for messages comparison.

### DIFF
--- a/Sources/CocoaLumberjack/DDLog.m
+++ b/Sources/CocoaLumberjack/DDLog.m
@@ -1052,7 +1052,7 @@ NS_INLINE BOOL _nullable_strings_equal(NSString* _Nullable lhs, NSString* _Nulla
     // See https://stackoverflow.com/questions/36593038/confused-about-the-default-isequal-and-hash-implements
     if (other == self) {
         return YES;
-    } else if (!other || ![other isKindOfClass:[self class]]) {
+    } else if (!other || ![other isKindOfClass:[DDLogMessage class]]) {
         return NO;
     } else {
         __auto_type otherMsg = (DDLogMessage *)other;
--- a/Sources/CocoaLumberjack/DDLog.m
+++ b/Sources/CocoaLumberjack/DDLog.m
@@ -1037,10 +1037,20 @@ NSString * __nullable DDExtractFileNameWithoutExtension(const char *filePath, BO
     return self;
 }
 
+static __inline__ __attribute__((__always_inline__)) BOOL _nullable_strings_equal(NSString* _Nullable lhs, NSString* _Nullable rhs)
+{
+    if (lhs == nil) {
+        if (rhs == nil)
+            return YES;
+    } else if (rhs != nil && [lhs isEqualToString:rhs])
+        return YES;
+    return NO;
+}
+
 - (BOOL)isEqual:(id)other {
     if (other == self) {
         return YES;
-    } else if (![super isEqual:other] || ![other isKindOfClass:[self class]]) {
+    } else if (!other || ![other isKindOfClass:[self class]]) {
         return NO;
     } else {
         __auto_type otherMsg = (DDLogMessage *)other;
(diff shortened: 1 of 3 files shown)


Sources/CocoaLumberjack/DDLog.m
```diff
@@ -1037,10 +1037,20 @@ - (instancetype)initWithMessage:(NSString *)message
...
+            return YES;
+    } else if (rhs != nil && [lhs isEqualToString:rhs])
+        return YES;
+    return NO;
+}
+
 - (BOOL)isEqual:(id)other {
     if (other == self) {
         return YES;
-    } else if (![super isEqual:other] || ![other isKindOfClass:[self class]]) {
```
ffried, line 1043: Can you add a comment why `super` isn't called here? Because overridden methods that do not call super is something that always catches my eye...

Sources/CocoaLumberjack/DDLog.m
```diff
@@ -1062,17 +1070,14 @@ - (BOOL)isEqual:(id)other {
 }
 
 - (NSUInteger)hash {
-    return [super hash]
```
ffried, line 1065: Same here. Please add a quick comment on why super isn't called. Although for hashing it could be included, right?

Sources/CocoaLumberjack/DDLog.m
ffried: Also, we could consider adding a `[super isEqual: other]` check right at the beginning. If super says we're equal, then we're equal. I'd still to the pointer check, in case `NSObject` ever decides to do something more.

Sources/CocoaLumberjack/include/CocoaLumberjack/DDLog.h
```diff
@@ -780,7 +780,7 @@ typedef NS_OPTIONS(NSInteger, DDLogMessageOptions){
     NSInteger _context;
     NSString *_file;
     NSString *_fileName;
-    NSString *_function;
+    NSString * _Nullable _function;
```
ffried, line 783: Not a review comment but a remark for release: This is a breaking change.

ffried (changes requested):

Sources/CocoaLumberjack/include/CocoaLumberjack/DDLog.h
jcbertin: It should not break anything since `_Nullable` is just an annotation: it could only generates warnings.

Sources/CocoaLumberjack/DDLog.m
jcbertin: See https://stackoverflow.com/questions/36593038/confused-about-the-default-isequal-and-hash-implements

Sources/CocoaLumberjack/DDLog.m
```diff
@@ -1037,10 +1037,22 @@ - (instancetype)initWithMessage:(NSString *)message
     return self;
 }
 
+static __inline__ __attribute__((__always_inline__)) BOOL _nullable_strings_equal(NSString* _Nullable lhs, NSString* _Nullable rhs)
```
ffried, line 1040: Foundation provides an NS_INLINE macro for `__inline__ __attribute__((__always_inline__))`

Sources/CocoaLumberjack/include/CocoaLumberjack/DDLog.h
jcbertin: Gotcha! removing it.

ffried (approved):

Issue #1276, opened 2022-01-04
Log rotate has some problem in 3.7.4 version


oyljerry:
### Issue Info

 Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform Name           |  osx
 Platform Version        | all
 CocoaLumberjack Version | 3.7.4
 Integration Method      | carthage
 Xcode Version           | Xcode 12.4
 Repro rate              | all the time (100%) 

### Issue Description and Steps

I have just upgraded CocoaLumberjack from 3.7.2 to 3.7.4, But I found there is some strange problems occur.

Below is code about how to control log rotate.

```
    fileLogger.maximumFileSize = 20 * 1024 * 1024;
    fileLogger.rollingFrequency = 0;
```

![image](https://user-images.githubusercontent.com/756387/148028901-b543163b-7b35-4301-a280-034a1d7ef620.png)

**Issuer#1:**

I found that the log is rotated when its size is only 1.3M, In 3.7.2, This works as 20M.

**Issuer#2:**

The log file which should be xxx-202112291336.log, but it has been changed to xxx-202212291336.log

**Is this issue related to https://github.com/CocoaLumberjack/CocoaLumberjack/pull/1234???**

ffried:
Are you using a subclass of `DDLogFileManagerDefault`?
If so, I could see both issues coming from an incomplete implementation:

The default log file name uses the following pattern as log file name: "<app name> <date> <time>.log"
Since your format differs from that, you've likely overridden `-newLogFileName`. To solve your date issue, please have a look at the date formatter you are using there. Specifically take care to use `yyyy` not `YYYY` for the year.

Also, if you override `-newLogFileName`, you also need to properly implement `-isLogFile:`! Otherwise the log files won't be recognized by other (non-overridden) implementations of `DDLogFileManagerDefault` and thus might not be reused correctly, resulting in new log files even though the criteria for rolling hasn't yet been met.

oyljerry:
Thanks @ffried  , The date format is wrong, I need to change from YYYY to yyyy.

oyljerry:
@ffried :
After some testing, We found **issuer1** still happened in below scenario. Please help to check it. 

1. On 2022-01-06, start app, the log generated date is aaaa-202201061900.log, 
2. I shutdown my macOS at 2022-01-06 22:00
3. After 1 day, at 2022-01-07, I power on my macOS, start my app.
[...]

PR #1287, opened 2022-02-11
Fix nullability violations

jcbertin:

### Pull Request Description

After enabling nullability violations checks, I found an issue where `_currentLogFileInfo` could be used before being initialized. This PR fixes that.


Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1325,6 +1328,8 @@ - (void)lt_logData:(NSData *)data {
     }
 
     @try {
+        NSFileHandle *handle = [self lt_currentLogFileHandle];
```
ffried, line 1331: Is there a specific reason for moving this up here?

ffried (approved):
Looks good. Just have one small question (see inline).

Sources/CocoaLumberjack/DDFileLogger.m
ffried: That might #1276

PR #1296, opened 2022-04-08
Support OSLogger on iOS 9.0

ssrlive:
Please approve it, I need it.


ffried:
Sorry, but we can't support OSLog on pre-iOS 10, because the necessary functions are only available as of iOS 10:
e.g. https://developer.apple.com/documentation/os/1643744-os_log_create
image

PR #1298, opened 2022-04-26
Update header file structure

amirthytejeshwar-sc:

* [ ] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
* [ ] I have read the [Documentation](http://cocoadocs.org/docsets/CocoaLumberjack/)
* [ ] I have searched for a similar pull request in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/pulls) and found none

* [ ] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)

### Pull Request Description
This pr resolves issue of `CocoaLumberjack.h not found`
So header files are placed in `Supporting Files` folder which was added to the target exclude parameter.
SPM works fine I think, it somehow recognises the headers or create a modulemap of it's own.
I had issues regarding the headers and tried fixing them. The headers should be present in `include` folder. Please CMIIW

Please review andsuggest changes if any required.
Also there is no `SwiftLogLevel.h` in `CocoaLumberjackSwift` but it exists in `CocoaLumberjackSwiftSupport`.


amirthytejeshwar-sc:
@ffried please review this

ffried:
I'll need to take a closer look at this. I know we purposely kept the Umbrella Header outside of SPMs reach (to let it generate its own).
Also, the SwiftSupport target is special. AFAIK it's a separate target in SPM, but included in `CocoaLumberjackSwift` in Xcode...

ffried:
In either way, the Xcode project needs adjusting (and probably the pod spec as well).

sushichop (commented):
Thanks for your PR.

Can you give us more information on when this issue occurs?
It would be helpful in investigating this issue.
 - How to integrate(CocoaPods, Carthage, SwiftPM, and manual...)
 - CocoaLumberjack's version
 - Sample project to repro (if possible)
 - ... 

Also, when you change the tree structure of files, you need to change the reference path to the file in Xcode. You can see that CI for `library-build` actually failed.

Thanks.

ffried (changes requested):
So, I just had a deeper look at this.

The setup was / is the following:

For SPM, we let SPM generate its own umbrella header and module map. I've been using this package like this in various projects and never had any issues. Thus, I'd be highly interested in a reproduction case that reliably reproduces any issues here.
This is why the headers are under the `Supporting Files` folder, that SPM is told to ignore, but Xcode will use.

The Xcode project (and CocoaPods spec) is a different story. Here we only have two targets: `CocoaLumberjack` and `CocoaLumberjackSwift`. The `CocoaLumberjackSwiftSupport` target of SPM is integrated in the Xcode target `CocoaLumberjackSwift`. That's the reason for the import in `CocoaLumberjackSwift.h` (which is ignored by SPM as stated above).

PR #1314, opened 2022-09-28
Don't log into symlink files

smaryus:

<br/>

- [ ] I have updated the documentation (if necessary)
- [ ] I have run the lint and it passes (`pod lib lint`)

### Pull Request Description

When the logger is trying to reuse an older log file, if the log file is a symlink the logger will write into the link file. This can result in different issues, specially if the log file names are custom. 
To prevent this, first check if the file is symlink, and when it is recreate it as regular file.

### DIFF
--- a/Sources/CocoaLumberjack/DDFileLogger.m
+++ b/Sources/CocoaLumberjack/DDFileLogger.m
@@ -1102,6 +1102,11 @@ NSTimeInterval     const kDDRollingLeeway              = 1.0;              // 1s
         return NO;
     }
 
+    // Don't followw symlink
+    if (logFileInfo.isSymlink) {
+        return NO;
+    }
+
     // If we're resuming, we need to check if the log file is allowed for reuse or needs to be archived.
     if (isResuming && (_doNotReuseLogFiles || [self lt_shouldLogFileBeArchived:logFileInfo])) {
         logFileInfo.isArchived = YES;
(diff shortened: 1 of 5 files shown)


Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1514,6 +1519,14 @@ - (NSTimeInterval)age {
     return -[[self creationDate] timeIntervalSinceNow];
 }
 
+- (BOOL)isSymlink {
+    if (self.fileAttributes[NSFileType] == NSFileTypeSymbolicLink) {
```
ffried, line 1523: This could be simplified to directly `return self.fileAttributes[NSFileType] == NSFileTypeSymbolicLink;`.

ffried (changes requested):
Thanks for this patch! Just two small changes - generally looks great!

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1102,6 +1102,11 @@ - (BOOL)lt_shouldUseLogFile:(nonnull DDLogFileInfo *)logFileInfo isResuming:(BOO
         return NO;
     }
 
+    // Don't followw symlink
```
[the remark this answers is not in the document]
smaryus: Done, thanks

Sources/CocoaLumberjack/DDFileLogger.m
smaryus: Done, thanks

ffried (approved):

gestas-projects:
Hello, when this fix will be released?

ffried:
I'd like to wait for macOS 13 and a usable Xcode 14 version to be released. Then we can ship a release that adds support for Swift 5.7, Xcode 14, etc. and the other changes in one go.

ffried:
@gestas-projects This is now released in version 3.8.0.

PR #1316, opened 2022-10-06
Update to Xcode 14 / Swift 5.7

ffried:
This requires dropping support for iOS/tvOS <11, macOS < 10.13, watchOS < 4 Also, Swift support < 5.5 is removed - SPM will automatically search for an older version.


ffried:
This requires a new minor version (3.8.0) since we're dropping support for some systems.

vishalSonawane:
When will this PR be merged? Is there any ETA for this?

ffried:
As soon as Xcode 14.1 is released.

vishalSonawane:
> As soon as Xcode 14.1 is released.

So can't we use it for Xcode 14 now? I am facing issues when I upgraded to Xcode 14 from Xcode 13.4.2.

The error says: 

`Module compiled with Swift 5.5.2 cannot be imported by the Swift 5.7 compiler`

ffried:
Xcode 14.0 doesn't include the macOS 13.0 SDK nor Swift 5.7 standard library. It's a bit a weird situation.

But that error should be fixable without this PR. It should simply need a clean build. You're trying to import a module that was built with Xcode <14 using Xcode 14...

vishalSonawane:
> Xcode 14.0 doesn't include the macOS 13.0 SDK nor Swift 5.7 standard library. It's a bit a weird situation.
> 
> But that error should be fixable without this PR. It should simply need a clean build. You're trying to import a module that was built with Xcode <14 using Xcode 14...

I have already tried this couple of times but no luck.

ffried:
@vishalSonawane What dependency management are you using? Maybe an Xcode clean build won't suffice if you are using e.g. Carthage which may pre-build binaries.

PR #1334, opened 2023-03-16
Fix for possible buffer overflow

williamgp:

- [X] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
- [X] I have read the [Documentation](http://cocoadocs.org/docsets/CocoaLumberjack/)
- [X] I have searched for a similar pull request in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/pulls) and found none

<br/>

- [X] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)
- [X] I have added the required tests to prove the fix/feature I am adding
- [X] I have updated the documentation (if necessary)
- [X] I have run the tests and they pass
- [X] I have run the lint and it passes (`pod lib lint`)

1. Changed `int` to `size_t` for `iovec_len` to avoid negative values and improve type safety.
2. Added `max_iovec_len` constant to ensure that the maximum size of the `v` array is known and to prevent buffer overflows.
3. Initialized `v` to zero to avoid potential use of uninitialized memory.
4. Removed unnecessary else branch for setting empty `iov_base` and `iov_len` for `v[0]`, `v[1]`, and `v[max_iovec_len - 1]`.
5. Added check to ensure that the length of the `msg` buffer does not exceed the maximum length allowed for the `iov_base` field in the `struct iovec` array.
6. Added check to ensure that `_automaticallyAppendNewlineForCustomFormatters` is true and that the last character of the `msg` buffer is not already a newline before adding a newline to `v[3]`.
7. Removed the `_automaticallyAppendNewlineForCustomFormatters` flag from the calculation of `iovec_len`, since it is now only used to determine whether to append a newline to `v[3]`.

Overall, these changes should help to improve the security of the code by preventing buffer overflows, reducing the risk of memory corruption, and improving type safety.

...

### DIFF
--- a/Sources/CocoaLumberjack/DDTTYLogger.m
+++ b/Sources/CocoaLumberjack/DDTTYLogger.m
@@ -1241,9 +1241,10 @@ static DDTTYLogger *sharedInstance;
         // Write the log message to STDERR
 
         if (isFormatted) {
-            // The log message has already been formatted.
-            const int iovec_len = (_automaticallyAppendNewlineForCustomFormatters) ? 5 : 4;
-            struct iovec v[iovec_len];
+        // The log message has already been formatted.
+            const size_t max_iovec_len = 5;
+            size_t iovec_len = (_automaticallyAppendNewlineForCustomFormatters) ? 5 : 4;
+            struct iovec v[max_iovec_len] = { 0 };
 
             if (colorProfile) {
                 v[0].iov_base = colorProfile->fgCode;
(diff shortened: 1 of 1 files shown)


Sources/CocoaLumberjack/DDTTYLogger.m
```diff
@@ -1241,9 +1241,10 @@ - (void)logMessage:(DDLogMessage *)logMessage {
         // Write the log message to STDERR
 
         if (isFormatted) {
-            // The log message has already been formatted.
-            const int iovec_len = (_automaticallyAppendNewlineForCustomFormatters) ? 5 : 4;
-            struct iovec v[iovec_len];
+        // The log message has already been formatted.
```
ffried, line 1244: NitPick: Indention seems wrong here.


```suggestion
           // The log message has already been formatted.
```

ffried (approved):
Thanks! I've added a small nitpick, but will fix it inline when merging.

PR #1351, opened 2023-07-26
Replace %s with modern Objective-C string creation

ffried:

<br/>

### Pull Request Description

[Apple's String Format docs](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Strings/Articles/formatSpecifiers.html#//apple_ref/doc/uid/TP40004265-SW1) recommend avoiding `%s` whenever possible:
image

I've also seen it cause issues when the underlying C-String has strange memory management. Using `@()` instead should solve that.

### DIFF
--- a/Sources/CocoaLumberjack/DDLog.m
+++ b/Sources/CocoaLumberjack/DDLog.m
@@ -974,13 +974,13 @@ NSString * __nullable DDExtractFileNameWithoutExtension(const char *filePath, BO
 #pragma clang diagnostic pop
 #endif
         _options      = options;
-        _timestamp    = timestamp ?: [NSDate new];
+        _timestamp    = timestamp ?: [NSDate date];
 
         __uint64_t tid;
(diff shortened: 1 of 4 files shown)


Sources/CocoaLumberjack/DDLog.m
```diff
@@ -974,13 +974,13 @@ - (instancetype)initWithFormat:(NSString *)messageFormat
...
         _options      = options;
-        _timestamp    = timestamp ?: [NSDate new];
+        _timestamp    = timestamp ?: [NSDate date];
 
         __uint64_t tid;
         if (pthread_threadid_np(NULL, &tid) == 0) {
             _threadID = [[NSString alloc] initWithFormat:@"%llu", tid];
         } else {
-            _threadID = @"missing threadId";
+            _threadID = @"N/A";
```
sushichop, line 983: I also think this is a better expression👍

sushichop (approved):

PR #1367, opened 2023-09-03
Use flock in file logger to coordinate writes with other processes

tmolitor-stud-tu:

<br/>

- [ ] I have added the required tests to prove the fix/feature I am adding
- [ ] I have updated the documentation (if necessary)
- [ ] I have run the tests and they pass
- [ ] I have run the lint and it passes (`pod lib lint`)

### Pull Request Description
This is a follow up patch I'm using for some time in Monal:
When using multiple processes logging to the same file (for example a Notification Service Extension, a Share Extension etc.), loglines sometimes are interleaved: the logline of one process is interrupted in the middle by the logline of another process, especially for long loglines.

We are using a "binary" log format in Monal (length prefixed json) which makes this problem more visible than with normal text logs (but this problem is present when using normal text logs, too).

@ffried If anything is missing here, just let me know! :)

### DIFF
--- a/Sources/CocoaLumberjack/DDFileLogger.m
+++ b/Sources/CocoaLumberjack/DDFileLogger.m
@@ -18,6 +18,9 @@
 #endif
 
 #import <sys/xattr.h>
+#import <sys/file.h>
+#import <errno.h>
+#import <unistd.h>
 
 #import "DDFileLogger+Internal.h"
 
@@ -1345,6 +1348,12 @@ static int exception_count = 0;
             [self willLogMessage:_currentLogFileInfo];
         }
 
+        // use an advisory lock to coordinate write with other processes
+        int fd = [handle fileDescriptor];
+        while(flock(fd, LOCK_EX) != 0) {
+            NSLogError(@"DDFileLogger: Could not lock logfile, retrying in 1ms: %s (%d)", strerror(errno), errno);
+            usleep(1000);
+        }
         if (@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)) {
             __autoreleasing NSError *error = nil;
             BOOL success = [handle seekToEndReturningOffset:nil error:&error];
@@ -1359,6 +1368,7 @@ static int exception_count = 0;
             [handle seekToEndOfFile];
             [handle writeData:data];
         }
+        flock(fd, LOCK_UN);
 
         if (implementsDeprecatedDidLog) {
 #pragma clang diagnostic push


tmolitor-stud-tu:
Sorry for the multiple force pushes, I've added some minor tweaks like the `DDFileLogger: ` in the error logging and the `#import <unistd.h>` (for me it works without `unistd.h`, but it's cleaner to explicitly import it, I think).

Sources/CocoaLumberjack/DDFileLogger.m
```diff
@@ -1345,6 +1348,12 @@ - (void)lt_logData:(NSData *)data {
             [self willLogMessage:_currentLogFileInfo];
         }
 
+        // use an advisory lock to coordinate write with other processes
+        int fd = [handle fileDescriptor];
+        while(flock(fd, LOCK_EX) != 0) {
```
ffried, line 1353: IMHO we should make this configurable in some way.

AFAIK the current default behavior of `DDFileLogger` will prevent this kind of issue by using a different file for different processes. 
So this only happens if a custom log file manager is used. So we could actually call out to the log file manager and either ask whether to lock the file, or even a general `willWriteToFile:` and `didWriteToFile:` callback, that in turn can be used by the log file manager to lock/unlock the file.

ffried (changes requested):
A good idea and patch in general. We just need to make sure it only affects relevant clients (see inline comment).

Sources/CocoaLumberjack/DDFileLogger.m
tmolitor-stud-tu: Well, if only one process ever accesses the logfile, the locking would be a noop and should not do any harm.

That said: I can, of course, implement it differently. Which one of your suggestions do you actually prefer?

tmolitor-stud-tu: @ffried I was thinking about this more:

Implementing `willWriteToFile:` and `didWriteToFile:` callbacks will leave the burden to remember to lock the file to the user writing the custom LogFileManager.
I'd prefer to either add an option to turn on/off the locking (ideally default on) or even better to just don't make this configurable at all.

Why that?
Well, that way _it just works_ out of the box for users that use the same logfile for multiple processes. And has virtually no impact on users that don't do this.

ffried (approved):

Sources/CocoaLumberjack/DDFileLogger.m
ffried: Yea, let's take it as is

Issue #1397, opened 2023-12-14
Privacy manifest for new sensitive APIs


nabs-m:
Apple recently announced privacy manifests, along with a list of [sensitive APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api) that require developers to provide reasons for them. Apple says devs have until spring 2024 to implement this.

I believe CocoaLumberjack uses:
* [NSFileCreationDate](https://developer.apple.com/documentation/foundation/nsfilecreationdate?language=objc) and [NSFileModificationDate](https://developer.apple.com/documentation/foundation/nsfilemodificationdate?language=objc) from the [File timestamp APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278393) list.
* statfs from the [Disk space APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278397) list.

Are you planning to include a privacy manifest describing this usage in a future release?

ffried:
I was already contacted by Apple regarding this. I'm currently waiting on feedback there.

nikekannan:
@ffried We are using CocoaLumberjack SDK, We are also in need of the same Privacy Manifest report.  Request your help and do the needful.

williamgp:
@ffried what do expect the correct required reason is for NSFileCreationDate/NSFileModificationDate?
Would C617.1 suffice?
> Declare this reason to access the timestamps, size, or other metadata of files inside the app container, app group container, or the app’s CloudKit container.

ffried:
See #1403 for details.

lamdaniel:
I have also double-checked this:
Can confirm usages of 
- NSFileCreationDate
- NSFileModificationDate
- statfs

PR #1403, opened 2024-01-09
Add privacy manifest

ffried:

<br/>

This merge request fixes / refers to the following issues: #1397 

### Pull Request Description

Adds a privacy manifest for the `CocoaLumberjack` target.

Open questions / points:

- [x] How will this integrate with CocoaPods?

### DIFF
--- a/Scripts/generate-podspec.sh
+++ b/Scripts/generate-podspec.sh
@@ -145,6 +145,7 @@ Pod::Spec.new do |s|
   s.subspec 'Core' do |ss|
     ss.source_files         = 'Sources/CocoaLumberjack/**/*.{h,m}'
     ss.private_header_files = 'Sources/CocoaLumberjack/DD*Internal.{h}'
+    ss.resources            = 'Sources/CocoaLumberjack/PrivacyInfo.xcprivacy'
   end
 
   s.subspec 'Swift' do |ss|
--- a/Scripts/generate-podspec.sh
+++ b/Scripts/generate-podspec.sh
@@ -77,6 +77,16 @@ IOS_SDK="$(read_config_var "${IOS_SDK_CONFIG_VAR}" '[0-9]+\.[0-9]+' "${SDKS_XCCO
 TVOS_SDK="$(read_config_var "${TVOS_SDK_CONFIG_VAR}" '[0-9]+\.[0-9]+' "${SDKS_XCCONFIG_FILE}")"
 WATCHOS_SDK="$(read_config_var "${WATCHOS_SDK_CONFIG_VAR}" '[0-9]+\.[0-9]+' "${SDKS_XCCONFIG_FILE}")"
 
+SUPPORTED_SWIFT_VERSIONS=''
+for SPM_PKG_DEF in Package@swift-*.swift; do
+    SWIFT_VERSION="$(echo "${SPM_PKG_DEF}" | sed -E 's/^Package@swift-([0-9]+\.[0-9]+)\.swift$/\1/g')"
+    if [[ -n "${SWIFT_VERSION}" ]] && [[ "${SWIFT_VERSION}" != "${SPM_PKG_DEF}" ]]; then
+        # We add the comma to the end here, since we will add the last version at the end.
+        SUPPORTED_SWIFT_VERSIONS="${SUPPORTED_SWIFT_VERSIONS}'${SWIFT_VERSION}', "
+    fi
+done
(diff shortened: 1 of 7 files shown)


nabs-m:
> How will this integrate with CocoaPods?

@ffried The privacy manifest (along with its path) needs to be added to the `podspec`'s `resources` attribute. I don't see a `podspec` file in this repo, but I'm guessing it would be added to [this script](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/f44b9cb1e3f9d6d2108476ea28d336c196e943a6/Scripts/generate-podspec.sh)? Probably something like this under [both subspecs](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/f44b9cb1e3f9d6d2108476ea28d336c196e943a6/Scripts/generate-podspec.sh#L145-L153):
```
ss.resources = 'Sources/CocoaLumberjack/PrivacyInfo.xcprivacy'
```

[Here's an example](https://github.com/bugsnag/bugsnag-cocoa/pull/1616/files#diff-c4c40d3e49de76d8fac0fa28120520e20f1131fbca42210efb854a3dde7bb105R56-R58) from Bugsnag that I know worked well for consuming apps where the privacy manifest was properly imported.

ffried:
@nabs-m Thanks a lot for this info! Saves me some research time!

Yea, we generate the pod spec on the fly to have one less file to manage.

I've added it to the script, though only to the Core subspec, since the Swift subspec has Core as its dependency. Thus I'd hope that this is picked up (works that way in SwiftPM). If not, I can always add it later on.

Scripts/generate-podspec.sh
```diff
@@ -77,6 +77,16 @@ IOS_SDK="$(read_config_var "${IOS_SDK_CONFIG_VAR}" '[0-9]+\.[0-9]+' "${SDKS_XCCO
...
+SUPPORTED_SWIFT_VERSIONS=''
+for SPM_PKG_DEF in Package@swift-*.swift; do
+    SWIFT_VERSION="$(echo "${SPM_PKG_DEF}" | sed -E 's/^Package@swift-([0-9]+\.[0-9]+)\.swift$/\1/g')"
+    if [[ -n "${SWIFT_VERSION}" ]] && [[ "${SWIFT_VERSION}" != "${SPM_PKG_DEF}" ]]; then
+        # We add the comma to the end here, since we will add the last version at the end.
+        SUPPORTED_SWIFT_VERSIONS="${SUPPORTED_SWIFT_VERSIONS}'${SWIFT_VERSION}', "
+    fi
+done
+SUPPORTED_SWIFT_VERSIONS="${SUPPORTED_SWIFT_VERSIONS}'$(swift package tools-version | awk -F'.' '{print $1"."$2}')'"
+
```
nabs-m, line 89: Nice 👌

nabs-m (approved):
Privacy manifest itself and the CocoaPods integration look good to me. I haven't tinkered with SPM integrations, but as long as it doesn't need to full path, it looks fine to me.

Scripts/generate-podspec.sh
ffried: Oh yeah. 🤠 
Don't know why I didn't to it like that in the first place...

ffried:
> I haven't tinkered with SPM integrations, but as long as it doesn't need to full path, it looks fine to me.

That's the nice thing about SPM: It'll (usually) tell you if it doesn't find something. 🙃 
The reason it's not needed her is that it already knows the path to the Sources. And since the manifest is in the root of the sources directory for this target, no (sub-)path is needed.

paulnicholson:
@ffried Thanks for your work on this! Is it ready to merge? How soon would it be released if it's merged? 
I don't think I can use a git reference for this in my podfile since the podspec is generated by a script.

ffried:
@paulnicholson I've passed this along to my contact at Apple (that originally reached out to me - as mentioned in #1397) and am waiting for feedback.
I'd expect the enforcement to happen some time in April. We should be able to release a version containing these changes until then.

layneios:
Hey team, sorry to tag here. Our app is using this CocoaLumberjack library, however, I got an error when generating Privacy Report. It shows me that missing an expected key: `NSPrivacyCollectedDataTypes` for this library. Could you make a fix if possible? 🙏

ffried:
That's strange. According to the documentation, the API types is the only key necessary for third-party SDKs: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files

image

ffried:
Interesting. I would have expected Xcode to verify the xcprivacy file of a target.
Either way, I added an empty list of collected data types, as well as `NSPrivacyTracking` with `NO`... That should do the trick.

As described in #1406, I'd release a new version towards the end of the week.

Issue #1409, opened 2024-02-13
DDOSLogger Info, Warning, Debug, and Verbose logs do not appear in Console.app when streaming from iOS simulator.


jaredsinclair:
## Checklist

- [x] I have read and understood the [CONTRIBUTING guide](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/.github/CONTRIBUTING.md)
- [x] I have read the [Documentation](https://github.com/CocoaLumberjack/CocoaLumberjack#documentation)
- [x] I have searched for a similar issue in the [project](https://github.com/CocoaLumberjack/CocoaLumberjack/issues) and found none

## Description

DDOSLogger Info, Warning, Debug, and Verbose logs do not appear in Console.app when streaming from iOS simulator.

## Steps To Reproduce

1. Implement a log message (via Swift or Objective-C, it doesn't matter) using DDOSLogger via any method that uses a DDLogFlag value of `.info`, `.debug`, .`warning`, or `.verbose`. 
2. Run that statement on an iOS or iPadOS simulator while streaming logs to Console.app on the same Mac.
3. These log messages do not appear even when Info and Debug logs are enabled in Console.app.

## Expected behavior

Every log message sent to CocoaLumberjack should appear in Console.app when Info and Debug logs are enabled in Console.app, even when streaming from the iOS simulator.

**Environment info**
  Info                    | Value                               |
-------------------------|-------------------------------------|
 Platform Name           | any Simulator device
 Platform Version        | at least as far back as iOS 16-17, probably before that, too
 CocoaLumberjack Version | latest
 Integration Method      | spm
 Xcode Version           | e.g. Xcode 15+
 Repro rate              | all the time
 Repro project link      | n/a

## Additional Context

Ultimately this is an Apple bug. The bug dates back [years and years](https://developer.apple.com/forums/thread/82736?page=2). However CocoaLumberjack should be aware of a longstanding bug like that and provide an automatic workaround that doesn't require end users to discover this shortcoming the hard way.

Specifically, [this switch statement in DDOSLogger](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/af4973721172dd7850a42a58a9ffcec0ec82e5a9/Sources/CocoaLumberjack/DDOSLogger.m#L105-L113) needs to be rethought:

```objc
switch (logMessage->_flag) {
    case DDLogFlagError  :
        os_log_error(logger, "%{public}s", msg);
        break;
[...]

PR #1410, opened 2024-02-14
Improve DDOSLogger

ffried:

<br/>

This merge request fixes / refers to the following issues: #1409 

### Pull Request Description

This improves the DDOSLogger implementation with points that were brought up in #1409.

- Cleanup properties and methods.
- Raise `DDLogFlagWarn` messages to the `.error` level in OSLog (which is what Apple does).
- Add `DDOSLogLevelMapper` to allow custom mapping from `DDLogFlag` -> `os_log_type_t`.

### DIFF
--- a/Sources/CocoaLumberjack/DDOSLogger.m
+++ b/Sources/CocoaLumberjack/DDOSLogger.m
@@ -13,18 +13,16 @@
 //   to endorse or promote products derived from this software without specific
 //   prior written permission of Deusty, LLC.
 
+#import <TargetConditionals.h>
 #import <os/log.h>
 
 #import <CocoaLumberjack/DDOSLogger.h>
 
-@interface DDOSLogger () {
-    NSString *_subsystem;
-    NSString *_category;
-}
+@interface DDOSLogger ()
 
-@property (copy, nonatomic, readonly, nullable) NSString *subsystem;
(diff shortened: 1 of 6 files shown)


Sources/CocoaLumberjack/DDOSLogger.m
```diff
@@ -13,105 +13,144 @@
...
+@implementation DDOSLogLevelMapperDefault
+
+- (instancetype)init {
+    self = [super init];
+    return self;
+}
+
+- (os_log_type_t)osLogTypeForLogFlag:(DDLogFlag)logFlag {
+    switch (logFlag) {
+        case DDLogFlagError:
```
lechuckcaptain, line 30: How about mapping `DDLogFlagError` to [OS_LOG_TYPE_FAULT](https://developer.apple.com/documentation/kernel/os_log_type_t/os_log_type_fault) ?

ffried: These kind of changes is exactly what the new level mapping facility can be used for.
I'd like to stick as close to the defaults here as possible, and since [`Logger.error` logs with the `error` level](https://developer.apple.com/documentation/os/logger/3551616-error), I'd do the same for our error log method.
As you correctly noted, to add `fault`, we'd need a level above `error`, which we currently don't have.

PR #1489, opened 2026-09-08
Bump minimum platform versions to the oldest Swift 6.4 supports

kdumaz:
## Problem

Swift 6.4 (Xcode 27) deprecates every platform declaration in this manifest, not just one. Raising the tools-version to 6.4 surfaces all four:

```
Package.swift:21:17: warning: 'v10_13' is deprecated: macOS 12.0 is the oldest supported version
Package.swift:22:15: warning: 'v12' is deprecated: iOS 15.0 is the oldest supported version
Package.swift:23:16: warning: 'v12' is deprecated: tvOS 15.0 is the oldest supported version
Package.swift:24:19: warning: 'v5' is deprecated: watchOS 9.0 is the oldest supported version
```

The deprecations are `@available(_PackageDescription, deprecated: 6.4, ...)`, so they are keyed to the manifest's declared **swift-tools-version**, not to the toolchain. At the current `swift-tools-version:6.3` a plain `swift build` on `master` stays quiet even under Xcode 27 — the warnings appear in Xcode's editor, and will appear for everyone the moment the tools-version moves to 6.4.

## Fix

Raise all four to the oldest version each platform still supports. Apple deprecated them as a set; fixing only `watchOS` leaves three identical warnings behind.

```diff
-        .macOS(.v10_13),
-        .iOS(.v12),
-        .tvOS(.v12),
-        .watchOS(.v5),
+        .macOS(.v12),
+        .iOS(.v15),
+        .tvOS(.v15),
+        .watchOS(.v9),
```

`swift-tools-version` is deliberately left at 6.3 so consumers are not forced onto a newer toolchain.

## Verification

Xcode 27.0 (27A5228h, Swift 6.4):

- before: 4 deprecation warnings with tools-version 6.4
- after: 0
- `swift build` completes cleanly at the manifest's current `swift-tools-version:6.3`

## Note on scope

This raises the package's minimum deployment targets, so it drops SwiftPM support for consumers below macOS 12 / iOS 15 / tvOS 15 / watchOS 9. That is a breaking change and may belong in a major release — happy to reduce this to `watchOS` only, or hold it for whichever release you would prefer.

## CI

`run-danger` is failing for a reason unrelated to this change: `danger/swift@3.22.1` crashes at startup with `ReferenceError: File is not defined` in `undici` under Node 18.20.8, before any rule is evaluated.


ffried:
@kdumaz Thank you for you PR.
As you correctly identified, this is a breaking change implemented like this. We'll raise these values with general Swift 6.4 support (which will maintain backwards compatibility with older Swift versions). That, in turn, will happen once Xcode 27 is live.

I don't think it makes sense to merge this as is, so I'm closing this for now and will open an issue for Swift 6.4 support you can track instead.

ffried:
See #1490

